PR #16, opened 2023-12-01
Add some basic docs

BrianKopp:


docs/examples/standard/README.md
```diff
@@ -0,0 +1,81 @@
...
+* S3 buckets
+* RabbitMQ Broker
+* Elasticache
+
+## Prerequisites
+
+For the standard install, you will need the following:
+
+* A Route53 Hosted Zone
+* The AWS ECR registry URL (get from Bigeye)
```
deeno35, line 23: I feel like we could just provide the ECR URL and put a comment in there that someone can change this if they want to cache our images?  Or are you worried about leaking our AWS account ID?

It would be super slick for us to update this dynamically, but perhaps not worth the effort.  Maybe this comment could be the aws CLI command on how to get the latest image tag.  Basically anything to make this self-service

deeno35: One thing they will need to do is contact our sales.  At a minimum SRE will need to authorize their AWS account ID for pulling images

docs/examples/standard/README.md
```diff
@@ -0,0 +1,81 @@
+# Standard Installation
```
deeno35, line 1: What do you think about examples as a top level dir.

/modules
/docs
/examples/standard

deeno35 (approved):

PR #17, opened 2023-12-01
Add script to automatically update tags in examples to latest

BrianKopp:


.github/workflows/test_and_publish.yaml
```diff
@@ -70,6 +70,8 @@ jobs:
           git-branch: "release-from-${{ github.sha }}"
           create-summary: true
           skip-git-pull: true
+      - name: replace versions in scripts
+        run: ./scripts/replace-version.sh "${{ steps.changelog.outputs.old }}" "${{ steps.changelog.outputs.tag }}"
```
deeno35, line 74: The tricky thing will be to get the app version number.  This is for the terraform-modules version number I think?

deeno35 (approved):

.github/workflows/test_and_publish.yaml
BrianKopp: Yes exactly. I'll update the script to say that

PR #20, opened 2023-12-05
Fix github action to get previous version

BrianKopp:


.github/workflows/test_and_publish.yaml
```diff
@@ -58,6 +58,11 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
+      - name: fetch current version
+        id: prevversion
+        run: echo "PREVIOUS_VERSION=v$(grep version package.json | awk '{print $2}' | tr -d '",')" >> "$GITHUB_OUTPUT"
```
deeno35, line 63: nit: `jq -r .version package.json` is on paper more deterministic, although I doubt we will ever end up with another field in there with version in the name

deeno35 (approved):

PR #21, opened 2023-12-05
Add documentation for self-managed-vpc

BrianKopp:


examples/self-managed-vpc/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+You will need all the same prerequisites as the standard
+install, as well as an existing VPC with appropriate subnets.
+
+## Configuration
+
+Follow all the configuration steps in the standard install,
+and then add the following variables:
+
+* `byovpc_vpc_id` - get this from the VPC you're installing within
```
deeno35, line 18: These are good comments, maybe these should also augment variables.tf and we only have the description in one place (variables.tf).  I'm a little torn about it, but in the end there will be drift if we have it in 2 places

BrianKopp: I think velocity is low enough that this is ok to be in both places.

examples/self-managed-vpc/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+Your VPC structure will dictate which subnet IDs need to be
+chosen for each category here. You must provide at least two
+subnet IDs for the variables that ask for the list. For example,
+AWS Application Load Balancers requires subnets from at least
+two availability zones.
+
+Your configuration might look something like:
+
+```hcl
```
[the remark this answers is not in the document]
BrianKopp: Sounds good!

PR #24, opened 2023-12-08
Port temporal worker env vars over from CF

deeno35:


modules/bigeye/variables.tf
```diff
@@ -687,6 +687,18 @@ variable "temporal_additional_secret_arns" {
   default     = {}
 }
 
+variable "temporal_per_namespace_worker_count" {
+  description = "Controls number of per-ns (scheduler, batcher, etc.) workers to run per namespace"
+  type        = number
+  default     = var.temporal_desired_count * 3
```
BrianKopp, line 693: Looks like this is unhappy about it

BrianKopp (commented):
Can you remove the package-lock.json changes? You may need to install your deps using `npm ci` instead of `npm install`

BrianKopp (approved):
Can you remove the package-lock.json changes? You may need to install your deps using `npm ci` instead of `npm install`

PR #42, opened 2023-12-21
fix: use bigeye as the default mysql db name for the app db

deeno35:
toro is part of a legacy naming convention that does not need to be carried into the future

BREAKING CHANGE: Existing installs will need to set the `datawatch_rds_db_name = 'toro'` variable or the upgrade will destroy the application database will all application settings (including users etc).

### DIFF
--- a/modules/bigeye/main.tf
+++ b/modules/bigeye/main.tf
@@ -1518,7 +1518,7 @@ module "datawatch_rds" {
   name   = "${local.name}-datawatch"
 
   # Connection Info
-  db_name                       = "toro"
+  db_name                       = var.datawatch_rds_db_name
   root_user_name                = "bigeye"
   root_user_password_secret_arn = local.datawatch_rds_password_secret_arn
   snapshot_identifier           = var.datawatch_rds_snapshot_identifier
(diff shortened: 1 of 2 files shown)


BrianKopp (approved):

modules/bigeye/main.tf
```diff
@@ -1518,7 +1518,7 @@ module "datawatch_rds" {
   name   = "${local.name}-datawatch"
 
   # Connection Info
-  db_name                       = "toro"
+  db_name                       = var.datawatch_rds_db_name
```
BrianKopp, line 1521: Is this a breaking change?

BrianKopp (dismissed):

modules/bigeye/main.tf
deeno35: Yes!  I added a BREAKING CHANGE comment with instructions on how to upgrade existing installs.  I think this will do the right thing with version number and the release note?

I don't think it's worth detailing how to upgrade without retaining, that involves downtime and manual intervention that in practice, only we would do internally

BrianKopp (approved):

PR #43, opened 2023-12-22
fix: monocle IAM Role could not use S3

BrianKopp:


deeno35 (approved):

modules/bigeye/main.tf
```diff
@@ -1093,11 +1093,17 @@ resource "aws_iam_role_policy" "monocle" {
       {
         Effect = "Allow"
         Action = [
-          "s3:GetObject",
           "s3:ListBucket",
-          "s3:PutObject"
         ]
         Resource = aws_s3_bucket.models.arn
+      },
```
deeno35, line 1099: For consistencies sake, we do have a Sid on these policy statemets for the large payload bucket.  We prob want to either have a sid for all or none

deeno35: Double checked, we could be more consistent with the name field in aws_iam_role_policy for the two buckets too.

PR #50, opened 2024-01-03
Add validation messages and fixes

BrianKopp:
Adds several conditions and error messages for validating that users provide the necessary data when they are managing the security groups, namely when `create_security_groups` is false.

Fixes several places where `create_security_groups=false` was not being respected.

Adds an environment variable to scheduler so it can talk to itself on the traffic port (default `80`).


jylertones (approved):
Looks good, clarified with Brian that _all of these vars_ should be set if `create_security_groups=false`, and that `datawatch_rds_db_name` has been set elsewhere to the correct db name to prevent breakage now that we're removing this hard-coded value.

PR #54, opened 2024-01-09
Create troubleshooting module

BrianKopp:


modules/bigeye/main.tf
```diff
@@ -711,7 +726,7 @@ module "haproxy" {
   vpc_cidr_block                = var.vpc_cidr_block
   subnet_ids                    = local.application_subnet_ids
   create_security_groups        = var.create_security_groups
-  additional_security_group_ids = var.haproxy_extra_security_group_ids
+  additional_security_group_ids = concat(var.haproxy_extra_security_group_ids, local.troubleshooting_instance_security_group_ids)
```
deeno35, line 729: Should we create a "client" SG for the ECS simple service so it all works similarly to how clients for the resources work?

deeno35 (approved):
I'm guessing this will change some since we talked about fargate for the admin container.

One thing that stands out is that it might be more consistent if we put a client SG into simpleservice that grants access since this is how access to other resources work.

Then I would expect the admin-container SG to be added to the various client SGs.

PR #58, opened 2024-01-11
De-sensitive variables that shouldn't be

BrianKopp:


modules/simpleservice/containers.tf
```diff
@@ -7,14 +7,21 @@ locals {
...
+    name   = var.name,
+    cpu    = local.container_cpu
+    memory = local.container_memory
+    image  = local.container_image
+    portMappings = [{
+      containerPort = var.traffic_port
+      hostPort      = var.traffic_port
+      protocol      = "tcp"
+    }]
+    essential    = true
```
deeno35, line 22: What does essential = true do?

deeno35 (approved):

modules/bigeye/variables.tf
```diff
@@ -231,13 +231,20 @@ variable "datadog_agent_memory" {
   default     = 512
 }
 
+# DEPRECATED - Use datadog_agent_api_key_secret_arn
```
deeno35, line 234: Are you going to just get this deployed to test-sre* then come back and remove this?  I'm also okay with just going fast and breaking things while we don't have a customer on here yet

modules/bigeye/locals.tf
```diff
@@ -119,18 +119,8 @@ locals {
...
-    DD_CALL_BASIC_CONFIG     = "false"
-    DD_TRACE_STARTUP_LOGS    = "true"
-    DD_TRACE_DEBUG           = "false"
-    DD_LOGS_INJECTION        = "true"
-    DATADOG_API_KEY          = var.datadog_agent_api_key
-  } : {}
-  toretto_dd_env_vars = var.datadog_agent_enabled ? {
-    DATADOG_API_KEY = var.datadog_agent_api_key
-  } : {}
+  create_datadog_secret            = var.datadog_agent_enabled && var.datadog_agent_api_key != "" && var.datadog_agent_api_key_secret_arn == ""
```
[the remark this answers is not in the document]
BrianKopp: I've added SRE-2679 and SRE-2678 for this breaking change. And added a label "TF Breaking Change" to linear for when we want to bundle up a handful of breaking changes.

modules/simpleservice/containers.tf
BrianKopp: It's an "essential" container as far as ECS is concerned. If it exit codes, it takes down the task.

>Suppose that the essential parameter of a container is marked as true, and that container fails or stops for any reason. Then, all other containers that are part of the task are stopped. If the essential parameter of a container is marked as false, then its failure doesn't affect the rest of the containers in a task. If this parameter is omitted, a container is assumed to be essential.

deeno35: I think we can omit it?  All of our containers are essential, ie we would want to term if the dd container or the main container die

PR #72, opened 2024-01-18
Pipe through admin container with environment variables

BrianKopp:


modules/troubleshooting/main.tf
```diff
@@ -10,6 +10,62 @@ terraform {
```
deeno35, line 10: I was thinking we should brand this thing as a feature and have everything (module directory name etc) be consistent all the way down.

bigeye-admin seemed pretty good.  Troubleshooting is good today, but I feel like we may do something more with this down the road

deeno35 (approved):

PR #83, opened 2024-01-24
Add rabbitmq plumbing and docs for admin container

BrianKopp:


docs/TROUBLESHOOTING.md
```diff
@@ -157,6 +157,15 @@ These two RDS databases are necessary for datawatch and temporal to start.
 This command tests that the databases are available, that they can listen
 on the correct port, and that login succeeds.
 
+### RabbitMQ
+
+```sh
+bigeye-admin test rabbit
```
deeno35, line 163: think we need to call this rabbitmq?   I like it the way it is now, but was thinking it might help some folks who are new (those tend to be the ones that read the docs)

deeno35 (approved):

docs/TROUBLESHOOTING.md
BrianKopp: Yep, I'll update that in upstream as well

PR #87, opened 2024-01-25
Add documentation for bring your own dns

BrianKopp:


examples/self-managed-dns/README.md
```diff
@@ -0,0 +1,106 @@
...
+| vanity_dns_name | haproxy_load_balancer_dns_name |
+| datawatch_dns_name | datawatch_load_balancer_dns_name |
+| datawatch_database_vanity_dns_name | datawatch_database_dns_name |
+| datawork_dns_name | datawork_load_balancer_dns_name |
+| metricwork_dns_name | metricwork_load_balancer_dns_name |
+| monocle_dns_name | monocle_load_balancer_dns_name  |
+| web_dns_name | web_load_balancer_dns_name  |
+| toretto_dns_name | toretto_load_balancer_dns_name  |
+| scheduler_dns_name | scheduler_load_balancer_dns_name  |
+| temporal_admin_dns_name | temporal_admin_load_balancer_dns_name  |
```
deeno35, line 96: should we stick to temporalui for the name of this service instead of temporal_admin?

examples/self-managed-dns/README.md
```diff
@@ -0,0 +1,106 @@
...
+| ------------------------------- | ----------------------------------- |
+| vanity_dns_name | haproxy_load_balancer_dns_name |
+| datawatch_dns_name | datawatch_load_balancer_dns_name |
+| datawatch_database_vanity_dns_name | datawatch_database_dns_name |
+| datawork_dns_name | datawork_load_balancer_dns_name |
+| metricwork_dns_name | metricwork_load_balancer_dns_name |
+| monocle_dns_name | monocle_load_balancer_dns_name  |
+| web_dns_name | web_load_balancer_dns_name  |
+| toretto_dns_name | toretto_load_balancer_dns_name  |
+| scheduler_dns_name | scheduler_load_balancer_dns_name  |
```
deeno35, line 95: Should we consider stripping this list down to just the bare minimums?  IE things that have an LB?  I'm just wondering if we do a ternary for the DB name for example (or just drop the DB alias altogether), if that makes life easier for someone doing the install.

examples/self-managed-dns/main.tf
```diff
@@ -0,0 +1,24 @@
...
+module "bigeye" {
+  source      = "git::https://github.com/bigeyedata/terraform-modules//modules/bigeye?ref=v1.12.0"
+  environment = local.environment
+  instance    = local.instance
+
+  top_level_dns_name  = local.top_level_dns_name
+  create_dns_records  = false
+  acm_certificate_arn = local.acm_certificate_arn
+
+  vpc_cidr_block = "10.241.0.0/16"
```
deeno35, line 17: Can prob leave the cidr block out of the example to keep this example as pure as possible

deeno35 (approved):

examples/self-managed-dns/README.md
BrianKopp: Sure! I think I based that on the DNS name we use, `-admin`. I'll change the output name

examples/self-managed-dns/main.tf
BrianKopp: Good idea!

examples/self-managed-dns/README.md
BrianKopp: Ok, cool. I've got those updated. I'll go through my example again tomorrow morning to make sure it's all good!

PR #95, opened 2024-01-31
Lots of misc fixes

BrianKopp:


modules/bigeye/locals.tf
```diff
@@ -143,8 +154,8 @@ locals {
   } : {}
   datawatch_dd_env_vars = var.datadog_agent_enabled ? {
     DD_TRACE_DISABLED_PLUGINS        = "dns"
-    DD_LOG_INJECTION                 = "true"
-    DD_TRACE_SAMPLE_RATE             = "1.00"
+    DD_LOGS_INJECTION                = "true"
+    DD_TRACE_SAMPLE_RATE             = "1.0"
```
deeno35, line 158: This one should be 1.00?  It probably just works either way I guess

modules/bigeye/main.tf
```diff
@@ -1305,6 +1371,7 @@ module "temporalui" {
   # Load balancer
   healthcheck_path                 = "/health"
```
deeno35, line 1372: Is this for temporalUI?  The health check path should be "/"

deeno35 (approved):

modules/bigeye/main.tf
BrianKopp: Yep it should be, updating...

PR #107, opened 2024-02-02
feat: set binlog format to ROW

deeno35:
default for RDS mysql is MIXED and this poses problems for most dataloaders that want to sync incrementally.  ROW is preferred.

### DIFF
--- a/modules/bigeye/main.tf
+++ b/modules/bigeye/main.tf
@@ -1967,40 +1967,35 @@ module "datawatch_rds" {
     {
       name  = "log_bin_trust_function_creators"
       value = "1"
-    },
-    {
+      }, {
       name  = "general_log"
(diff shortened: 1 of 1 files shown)


modules/bigeye/main.tf
```diff
@@ -1967,40 +1967,35 @@ module "datawatch_rds" {
...
       name  = "innodb_lock_wait_timeout"
       value = 300
-    },
-    {
+      }, {
       name  = "lock_wait_timeout"
       value = 300
+      }, {
+      name  = "binlog_format"
+      value = "ROW"
```
deeno35, line 1998: This is the change here - added binlog_format=ROW.  The rest are white space changes

BrianKopp (approved):

PR #125, opened 2024-02-14
feat: add control over temporal persistence QPS

deeno35:
These knobs are useful to control throughput on the Temporal DB. https://docs.temporal.io/references/dynamic-configuration

### DIFF
--- a/modules/bigeye/variables.tf
+++ b/modules/bigeye/variables.tf
@@ -902,6 +902,42 @@ variable "temporal_logging_enabled" {
   default     = true
 }
 
+variable "temporal_frontend_persistence_max_qps" {
+  description = "Maximum number queries per second that the Frontend Service host can send to the Persistence store."
+  type        = string
+  default     = 2000
+}
+
+variable "temporal_history_persistence_max_qps" {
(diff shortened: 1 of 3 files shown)


modules/bigeye/variables.tf
```diff
@@ -902,6 +902,42 @@ variable "temporal_logging_enabled" {
   default     = true
 }
 
+variable "temporal_frontend_persistence_max_qps" {
+  description = "Maximum number queries per second that the Frontend Service host can send to the Persistence store."
+  type        = string
```
BrianKopp, line 907: string or number?

BrianKopp (approved):

modules/bigeye/variables.tf
deeno35: Good call, I'll update it

deeno35: Error: ECS Task Definition container_definitions is invalid: decoding JSON: json: cannot unmarshal number into Go struct field KeyValuePair.Environment.Value of type string

I was wondering what would happen here.  Now we know!

PR #129, opened 2024-02-16
Make the 4 email server params configurable

deeno35:
commit da39491e3063ece27557b7dd429571a488953191 (HEAD -> matthalverson/one-1339-code-change-make-the-4-email-server-params-configurable, origin/matthalverson/one-1339-code-change-make-the-4-email-server-params-configurable)
Date:   Thu Feb 15 18:34:38 2024 -0800

    docs: add documentation for byo mail server

commit d90425a9a706e27ab7a4900672b8425913db9d38
Date:   Thu Feb 15 18:12:45 2024 -0800

    feat: add BYO mail server env vars to datawatch, datawork, metricwork

    These vars are for plumbing in a custom SMTP server.  The app is
    set up to fall back to default for both empty string env vars
    and non-existant env vars.

    I opted to only enable the feature if all 4 required env vars are
    specified to avoid any kind of mistakes when trying to use the feature

### DIFF
--- a/modules/bigeye/locals.tf
+++ b/modules/bigeye/locals.tf
@@ -115,7 +115,7 @@ locals {
   sentry_event_level_env_variable = var.sentry_event_level != "" ? {
     SENTRY_EVENT_LEVEL = var.sentry_event_level
   } : {}
-  sentry_dsn_secret_arn = var.sentry_dsn_secret_arn != "" ? {
+  sentry_dsn_secret_map = var.sentry_dsn_secret_arn != "" ? {
     SENTRY_DSN = var.sentry_dsn_secret_arn
   } : {}
 
@@ -129,7 +129,7 @@ locals {
     local.auth0_secrets_map,
     local.slack_secrets_map,
     local.stitch_secrets_map,
-    local.sentry_dsn_secret_arn,
+    local.sentry_dsn_secret_map,
     local.byomailserver_smtp_password_secrets_map,
     var.datawatch_additional_secret_arns,
     {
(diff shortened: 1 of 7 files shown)


modules/bigeye/variables.tf
```diff
@@ -394,6 +394,31 @@ variable "stitch_api_token_secretsmanager_arn" {
...
 }
 
+variable "byomailserver_smtp_host" {
+  description = "Hostname of SMTP server.  This is for routing email notifications through your customer SMTP server vs using Bigeye's.  Ex. smtp.example.com"
+  type        = string
+  default     = ""
+}
+
+variable "byomailserver_smtp_port" {
+  description = "Port for the SMTP server.  This is for routing email notifications through your customer SMTP server vs using Bigeye's."
```
BrianKopp, line 404: should be number? Default to 25?

modules/bigeye/variables.tf
```diff
@@ -394,6 +394,31 @@ variable "stitch_api_token_secretsmanager_arn" {
...
+  default     = ""
+}
+
+variable "byomailserver_smtp_user" {
+  description = "SMTP credentials for your custom SMPT server."
+  type        = string
+  default     = ""
+}
+
+variable "byomailserver_smtp_password_arn" {
```
BrianKopp, line 415: nit: byomailserver_smtp_password_secret_arn is how we've been naming the secret arns

modules/bigeye/locals.tf
```diff
@@ -124,6 +130,7 @@ locals {
     local.slack_secrets_map,
     local.stitch_secrets_map,
     local.sentry_dsn_secret_arn,
+    local.byomailserver_smtp_password_arn,
```
BrianKopp, line 133: nit: byomailserver_smtp_password_secrets_map and can you change sentry_dsn_secret_map as well? I named that poorly to begin with. Existing name would suggest this is just a string not a map

BrianKopp (approved):

modules/bigeye/variables.tf
deeno35: Oh good call thanks

modules/bigeye/variables.tf
deeno35: env vars all have to be strings because of how it's getting expanded as an ECS config :(  I tried it out with a number before in a different PR and it and blew up nicely on me

deeno35: I also opted for no default.  I don't think we want to allow people to override part of our default, IE use a different port number, but still try to connect to our default mail server so I made the settings all or none.

PR #136, opened 2024-02-22
feat: add ability to set ingress cidr blocks for rabbitmq

deeno35:
The usual method of allowing attachment of additional security groups to grant access does not work for Amazon MQ after MQ creation.  To side step this and still allow addtional access to Rabbit's AMQPS and HTTPS admin console, this PR allows passing in extra cidr blocks directly.

### DIFF
--- a/modules/rabbitmq/main.tf
+++ b/modules/rabbitmq/main.tf
@@ -66,10 +66,27 @@ resource "aws_security_group" "this" {
   vpc_id = var.vpc_id
 
   ingress {
+    description     = "AMPQS connections from Bigeye"
     from_port       = 5671
     to_port         = 5671
     protocol        = "TCP"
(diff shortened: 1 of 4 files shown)


modules/rabbitmq/main.tf
```diff
@@ -72,6 +72,21 @@ resource "aws_security_group" "this" {
     security_groups = [aws_security_group.client[0].id]
   }
 
+  ingress {
+    from_port   = 5671
+    to_port     = 5671
+    protocol    = "TCP"
+    cidr_blocks = var.extra_ingress_cidr_blocks
```
BrianKopp, line 79: nit: can you throw a description on these please?

BrianKopp (approved):

PR #148, opened 2024-02-29
ci: rename SDP github action

deeno35:
The workflow that this GH action is calling has been renamed.


BrianKopp (approved):
nit: can you reword the commit to use `ci:`

PR #155, opened 2024-03-01
feat: add variable to control from email for byo smtp server setups

deeno35:
For installs where a custom SMTP server is being used, we want to
    require setting the "from address" to a domain that the installer
    controls, ie the mail server is verified to send for (dkim, spf, dmarc.)

### DIFF
--- a/modules/bigeye/locals.tf
+++ b/modules/bigeye/locals.tf
@@ -52,10 +52,11 @@ locals {
   create_adminpages_password_secret    = var.adminpages_password_secret_arn == ""
   adminpages_password_secret_arn       = local.create_adminpages_password_secret ? aws_secretsmanager_secret.adminpages_password[0].arn : var.adminpages_password_secret_arn
   # byomailserver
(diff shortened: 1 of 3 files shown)


modules/bigeye/locals.tf
```diff
@@ -52,10 +52,11 @@ locals {
   create_adminpages_password_secret    = var.adminpages_password_secret_arn == ""
   adminpages_password_secret_arn       = local.create_adminpages_password_secret ? aws_secretsmanager_secret.adminpages_password[0].arn : var.adminpages_password_secret_arn
   # byomailserver
-  byomailserver_enabled                   = var.byomailserver_smtp_host != "" && var.byomailserver_smtp_port != "" && var.byomailserver_smtp_user != "" && var.byomailserver_smtp_password_secret_arn != ""
+  byomailserver_enabled                   = var.byomailserver_smtp_host != "" && var.byomailserver_smtp_port != "" && var.byomailserver_smtp_user != "" && var.byomailserver_smtp_password_secret_arn != "" && var.byomailserver_smtp_from_address != ""
```
deeno35, line 55: All or nothing still for byo mail server.  It doesn't make sense for someone's install to try and send email as @bigeye.com to their custom mailserver and similarly it doesn't make sense to use Bigeye's mail server to send email from a custom domain.

BrianKopp (approved):

PR #164, opened 2024-03-05
feat: add ability to specify models s3 bucket name

BrianKopp:


modules/bigeye/variables.tf
```diff
@@ -518,6 +518,12 @@ variable "monocle_autoscaling_request_count_target" {
   default     = 15
 }
 
+variable "monocle_models_s3_bucket_name_override" {
```
deeno35, line 521: Should we name this around our "byo" naming convention?

deeno35 (approved):

PR #167, opened 2024-03-06
docs: update self-managed-vpc-no-internet-access instructions

deeno35:
There are tons of new features now, update this set of instructions with a working end to end install.  Normal installs will only require what's in main.tf, but for the purpose of creating a functional end to end install, all of the addtional *_example.tf files have been added to create addtional resources necessary such as SES, Workmail, a bastion for SSH tunnel, etc.

### DIFF
--- a/examples/self-managed-vpc-no-internet-access/README.md
+++ b/examples/self-managed-vpc-no-internet-access/README.md
@@ -1,53 +1,221 @@
-# Self-Managed VPC
+# Self-managed VPC with no internet access installation
 
 Some installation environments may require installing Bigeye in a VPC that has no NAT or route for
-any services or resources to public net. The method for doing this is to create your own (or bring 
-an existing) VPC that does not have internet access and pass it into the Bigeye module.  
+any services or resources to public net. The method for doing this is to create your own (or bring
+an existing) VPC that does not have internet access and pass it into the Bigeye module.
 
-This example shows the necessary configuration of variables for the Bigeye module to accomplish 
+The only file necessary to install Bigeye with existing infrastructure (DNS domain, vpc, security groups,
+mail server, etc) is [main.tf](./main.tf).  All of the other `*_example.tf` files are here as
+a reference as they will create example infrastructure (DNS domain, vpc, security groups,
+mail server, etc) for an end to end test.
+
+This example shows the necessary configuration of variables for the Bigeye module to accomplish
 this as well as a sample VPC configuration with the necessary VPC Endpoints (S3, logs, ECR) and
 security groups that can be used as a reference.  The VPC endpoints are required for the services
 to function properly without access to AWS' public service endpoints and is just overall a recommended
 practice, even if your installation will have access to/from public net.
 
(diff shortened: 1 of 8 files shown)


examples/self-managed-vpc-no-internet-access/README.md
```diff
@@ -1,53 +1,211 @@
...
+# Self-managed VPC with no internet access installation
 
 Some installation environments may require installing Bigeye in a VPC that has no NAT or route for
-any services or resources to public net. The method for doing this is to create your own (or bring 
-an existing) VPC that does not have internet access and pass it into the Bigeye module.  
+any services or resources to public net. The method for doing this is to create your own (or bring
+an existing) VPC that does not have internet access and pass it into the Bigeye module.
 
-This example shows the necessary configuration of variables for the Bigeye module to accomplish 
+The only file necessary is to install Bigeye with existing infrastructure (DNS domain, vpc, security groups,
```
BrianKopp, line 7: nit: should this say `The only file necessary to ...`

examples/self-managed-vpc-no-internet-access/main.tf
```diff
@@ -1,42 +1,113 @@
...
+  # The bastion will be used for an SSH tunnel for no internet installs where a VPN is not available.  This is here for
+  # demonstration purposes in this example, but a VPN is recommended instead for production installs.
+  bastion_enabled = false
+
+  # If you plan to use the bastion, run ssh-keygen -t rsa -b 4096 to create a key pair first if you have not already
+  bastion_ssh_public_key_file = "~/.ssh/id_rsa.pub"
+
+  # https://whatsmyip.com can be used to get your IP if you don't have it.  This ingres cidr goes into the
+  # security group allowing access to the bastion.
+  bastion_ingress_cidr = "24.7.117.81/32"
```
BrianKopp, line 71: Is this anonymized, or does that matter?

examples/self-managed-vpc-no-internet-access/outputs.tf
```diff
@@ -0,0 +1,15 @@
...
+output "bigeye_first_time_account_creation_address" {
+  value = "https://${local.vanity_alias}.${local.subdomain}/first-time"
+}
+
+output "bastion" {
+  value = one(aws_instance.bastion[*].public_dns)
+}
+
+output "bastion_ssh_user" {
+  value = "ssh ubuntu@<bastion>"
```
BrianKopp, line 14: Should we be interpolating the IP here?

BrianKopp (approved):

examples/self-managed-vpc-no-internet-access/main.tf
deeno35: Yes, thanks!

examples/self-managed-vpc-no-internet-access/outputs.tf
deeno35: It's one of those tricky things about condition responses in output.  I could put it in there, but then it looks strange when the flag is off for the bastion and ends up with this:

`ssh ubuntu@`

I'll see if I can improve this

PR #173, opened 2024-03-08
docs: restrict IPs to local subnet in no-internet example

deeno35:
This currently relies on the networking routes/subnets to prevent external access.  We should also be restricting the security groups to only allow ingress/egress from the VPC cidr in this example to add a 2nd layer of protection to the security moat.

### DIFF
--- a/examples/self-managed-vpc-no-internet-access/vpc_example.tf
+++ b/examples/self-managed-vpc-no-internet-access/vpc_example.tf
@@ -112,12 +112,11 @@ resource "aws_security_group" "vpc_endpoint" {
   }
 
   egress {
-    from_port        = 0
-    to_port          = 65535
-    protocol         = "TCP"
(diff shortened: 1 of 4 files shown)


examples/self-managed-vpc-no-internet-access/vpc_example.tf
```diff
@@ -112,12 +112,11 @@ resource "aws_security_group" "vpc_endpoint" {
   }
 
   egress {
-    from_port        = 0
-    to_port          = 65535
-    protocol         = "TCP"
-    cidr_blocks      = ["0.0.0.0/0"]
-    ipv6_cidr_blocks = ["::/0"]
```
deeno35, line 119: vpc is set to enable_ipv6=false anyways so also ditching the ipv6 rules so there is less for installers to review

ajribeiro (approved):

BrianKopp (approved):

PR #180, opened 2024-03-14
feat: create parameter group by default for datawatch replica

BrianKopp:


modules/bigeye/variables.tf
```diff
@@ -1139,42 +1179,28 @@ variable "datawatch_rds_replica_backup_retention_period" {
...
-      }, {
-      name  = "character_set_server"
-      value = "utf8mb4"
+      name         = "binlog_row_image"
+      value        = "full"
+      apply_method = "immediate"
       }, {
-      name  = "general_log"
-      value = 0
+      name  = "enforce_gtid_consistency"
```
deeno35, line 1191: This used to break the world if changed after a replica is seeded.  The docs say it should work now, not sure, but we should probably try it first if we haven't already.  "Self-hosted" instructions involve running stop replica etc.  Maybe AWS has handled it for us, not sure.

modules/bigeye/variables.tf
```diff
@@ -1139,42 +1179,28 @@ variable "datawatch_rds_replica_backup_retention_period" {
...
+      value        = "0"
+      apply_method = "immediate"
       }, {
-      name  = "lock_wait_timeout"
-      value = 300
       name  = "log_bin_trust_function_creators"
       value = "1"
       }, {
-      name  = "long_query_time"
-      value = 120
```
deeno35, line 1177: Whenever/if ever we turn on slow logging, I don't think it's worth logging queries that take longer than 10s (mysql default) even on the replicas.  Probably users will want to configure this number.  The internal monitoring queries we use run quite long as "normal" due to joins and table size.

modules/bigeye/variables.tf
```diff
@@ -1139,42 +1179,28 @@ variable "datawatch_rds_replica_backup_retention_period" {
...
+  description = "Parameters to use for the RDS replica database"
+  type        = list(map(string))
   default = [
     {
-      name  = "binlog_format"
-      value = "ROW"
-      }, {
-      name  = "character_set_server"
-      value = "utf8mb4"
+      name         = "binlog_row_image"
```
deeno35, line 1187: binlog_row_image full is the default so we can prob omit this.

binlog_format ROW is required for stitch

modules/bigeye/variables.tf
```diff
@@ -1139,42 +1179,28 @@ variable "datawatch_rds_replica_backup_retention_period" {
...
-variable "datawatch_rds_additional_tags" {
-  description = "Additional tags to apply to the datawatch RDS resources"
-  type        = map(string)
-  default     = {}
-}
-
-variable "datawatch_rds_parameters" {
-  description = "Parameters to use for the RDS database. See https://github.com/terraform-aws-modules/terraform-aws-rds?tab=readme-ov-file#input_parameters"
-  type        = list(map(any))
+variable "datawatch_rds_replica_parameters" {
```
deeno35, line 1182: Should we allow users to pass in overrides instead of them having to specify the entire set all or nothing?

deeno35 (approved):

modules/bigeye/variables.tf
BrianKopp: Yep, you're right about binlog_row_image!

modules/bigeye/variables.tf
BrianKopp: this diff here is a lie. It got moved up since my variables were out of order. The setting is retained by default for DW db params

PR #194, opened 2024-04-04
fix: add sentry dsn to web container

BrianKopp:


jylertones (approved):
I don't understand how this links up to `SENTRY_DSN`, but this looks like it's better than what we have today

PR #203, opened 2024-04-09
feat: split temporal into components

BrianKopp:
Fixes SRE-2779
Fixes SRE-3123


modules/bigeye/main.tf
```diff
@@ -1068,433 +1068,6 @@ module "temporal_rds" {
   replica_additional_tags               = var.temporal_rds_primary_additional_tags
 }
 
-resource "aws_security_group" "temporal_lb" {
```
BrianKopp, line 1071: Moved all of this over into `temporal.tf`

modules/bigeye/temporal.tf
```diff
@@ -0,0 +1,654 @@
+#======================================================
+# Frontend Service
+#
+# The temporal frontend service is responsible for
+# receiving external requests and processing them
+# in the cluster
+#======================================================
+resource "aws_lb" "temporal" {
+  name                             = "${local.name}-temporal"
```
BrianKopp, line 9: Question on these. We could change the name of the existing resources to `-temporal-frontend` and then mark this as a breaking change since it'll take down the temporal load balancer and have a DNS swap. Or, we could leave the frontend service as `-temporal`. We could also pass an option to NOT separate the temporal services, and then the `temporal_components` resources below would be conditional.

modules/bigeye/temporal.tf
```diff
@@ -0,0 +1,667 @@
...
+      dockerLabels = merge(local.temporal_datadog_docker_labels_generic, {
+        "com.datadoghq.ad.instances" : "[\n    {\n      \"openmetrics_endpoint\": \"http://localhost:9091/metrics\",\n      \"collect_histogram_buckets\": true,\n      \"histogram_buckets_as_distributions\": true,\n      \"collect_counters_with_distributions\": true,\n      \"tags\": [\n        \"app:temporal\",\n        \"component:frontend\"\n        \"instance:${var.instance}\",\n        \"stack:${local.name}\"\n      ]\n    }\n  ]\n",
+        "com.datadog.tags.app" : "temporal"
+        "com.datadog.tags.component" : "frontend"
+        "com.datadog.tags.service" : "temporal-frontend"
+      })
+    },
+  )
+
+  temporal_component_datadog_container_def = {
```
deeno35, line 581: Would be nice if the frontend one above could fit into this one (DRY)

modules/bigeye/variables.tf
```diff
@@ -925,14 +925,47 @@ variable "temporal_rds_replica_additional_tags" {
...
+# Application Variables - Temporal Components
+#======================================================
 variable "temporal_desired_count" {
-  description = "The desired number of replicas"
+  description = "The desired number of temporal components, applies to frontend, internal_frontend, history, matching, and worker services if those are not set"
   type        = number
   default     = 1
 }
 
+variable "temporal_desired_count_frontend" {
```
deeno35, line 937: `'<service>_<component>_<setting>'` reads better to me for all of these.  ie temporal_frontend_desired_count

deeno35 (approved):

deeno35 (approved):

PR #210, opened 2024-04-23
feat: add encryption and salt secrets

BrianKopp:


modules/bigeye/main.tf
```diff
@@ -2212,6 +2212,40 @@ resource "aws_secretsmanager_secret_version" "robot_password" {
   secret_string = random_password.robot_password[0].result
 }
 
+resource "random_password" "base_encryption" {
+  count   = local.create_base_dw_encryption_secret ? 1 : 0
+  length  = 16
```
szhongren, line 2217: should we have this be longer?

BrianKopp: Yes, good catch thank you

modules/bigeye/locals.tf
```diff
@@ -151,6 +155,8 @@ locals {
       MQ_BROKER_PASSWORD     = local.rabbitmq_user_password_secret_arn
       MYSQL_PASSWORD         = local.datawatch_rds_password_secret_arn
       ROBOT_PASSWORD         = local.robot_password_secret_arn
+      BASE_ENCRYPTION_STRING = local.base_datawatch_encryption_secret_arn
```
szhongren, line 158: should change to BASE_ENCRYPTION_SECRET

modules/bigeye/variables.tf
```diff
@@ -1059,6 +1059,18 @@ variable "datawatch_robot_password_secret_arn" {
   default     = ""
 }
 
+variable "datawatch_base_encryption_secret_arn" {
+  description = "ARN for secretsmanager secret holding the base encryption secret. One will be created if not provided."
```
deeno35, line 1063: Maybe 1 more sentence in here (and the salt) for what the encryption secret and salt is used for will probably be helpful

deeno35 (approved):

modules/bigeye/variables.tf
szhongren: They are used together to generate the AES symmetric key to actually do the encryption and decryption with

szhongren: Generation is done in Java code

BrianKopp: How's that look?

PR #225, opened 2024-05-03
fix: fix db parameter apply method not working

deeno35:
Small bug here which had a side effect of TF plans always looking like they need to change the apply_method on certain parameters.


modules/rds/main.tf
```diff
@@ -37,12 +37,11 @@ locals {
     var.replica_parameters
   )
   replica_parameters_list = [
-    for k, v in local.replica_parameters_object : merge(
-      {
-        name = k
-      },
-      v
-    )
+    for k, v in local.replica_parameters_object : {
```
deeno35, line 40: This actually is functionally equivalent, but I didn't see a reason why we should process replica params differently than the primary params, so copy pasted to make them look the same.

modules/rds/variables.tf
```diff
@@ -217,8 +217,8 @@ variable "replica_create_parameter_group" {
 variable "replica_parameters" {
   description = "A map of parameters, whose keys are the parameter name, value is the parameter value, and apply_type is optionally provided as pending-reboot or immediate"
   type = map(object({
-    value      = any
-    apply_type = optional(string)
+    value        = any
+    apply_method = optional(string)
```
deeno35, line 221: Optionals can lead to weird behavior since TF surprisingly was ignoring what an undefined field being passed in 😢

BrianKopp: Good catch. I thought I had changed this

BrianKopp (approved):

PR #235, opened 2024-05-10
Split out lineagework service

deeno35:
commit 9d8a03c784deddc42ea38acff37d24b01300bc0b (HEAD -> david/sre-3258-split-out-lineagework-service, origin/david/sre-3258-split-out-lineagework-service)
Date:   Fri May 10 12:24:44 2024 -0700

    feat: add lineagework service

    This is splitting out the lineage background processes into its
    own dedicated set of workers to improve service/job isolation and
    improve visibility.

    While not a breaking change, installs that do not have lineage plus
    installed can just set the lineagework instance count to 0 to save
    on costs as that service only pertains to work for the lineage plus
    feature.  Installs with lineageplus will want to tune cpu/mem
    for both the datawork and lineagework services after deployment.
    Setting CPU/mem for lineagework to the same as datawork is a reasonable
    starting point.

commit 7ade43adabf4526e21654c48740d2861d0b18fd2
Date:   Fri May 10 12:23:45 2024 -0700

    style: git ignore vscode configs

    Don't check in IDE preferences


modules/bigeye/main.tf
```diff
@@ -2062,14 +2085,83 @@ module "datawork" {
...
+  # lineagework should only subscribe to source-lineage and metacenter-lineage, ie exclude all other Temporal queues.
+  environment_variables = merge(
+    local.datawatch_dd_env_vars,
+    local.datawatch_common_env_vars,
+    {
+      APP                = "lineagework"
+      DATAWATCH_ADDRESS  = "http://localhost:${var.lineagework_port}"
+      WORKERS_ENABLED    = "true"
+      METRIC_RUN_WORKERS = "1"
+      EXCLUDE_QUEUES     = "run-metrics.v1,delete-source.v1,get-samples.v1,collect-lineage.v1,indexing.v1,reconciliation,trigger-batch-metric-run,agent-heartbeat"
```
deeno35, line 2157: After a chat internally, we decided to implement subscribing to "2 queues" this way for now vs making a code change to handle that case since this area is still under development and will change over the coming months.

BrianKopp (approved):

PR #240, opened 2024-05-13
Add elasticsearch to terraform module

BrianKopp:


modules/bigeye/variables.tf
```diff
@@ -972,6 +972,60 @@ variable "temporal_rds_replica_additional_tags" {
...
+variable "temporal_opensearch_engine_version" {
+  description = "The opensearch engine version"
+  type        = string
+  default     = "OpenSearch_2.11"
+}
+
+variable "temporal_opensearch_instance_type" {
+  description = "The opensearch instance type to use"
+  type        = string
+  default     = "m6g.large.search"
```
deeno35, line 990: Should we use something small for the base install (ie t3.medium.search)?

deeno35 (approved):

modules/bigeye/variables.tf
BrianKopp: Good call! I'll update that

PR #250, opened 2024-05-15
feat: allow passing in rds db option group settings

deeno35:
This is useful for enabling things such as mysql audit logging.

### DIFF
--- a/modules/bigeye/variables.tf
+++ b/modules/bigeye/variables.tf
@@ -937,6 +937,18 @@ variable "temporal_rds_additional_tags" {
   default     = {}
 }
 
+variable "temporal_rds_options" {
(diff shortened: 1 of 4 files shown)


modules/bigeye/variables.tf
```diff
@@ -1380,6 +1392,18 @@ variable "datawatch_rds_enabled_logs" {
   default     = ["error"]
 }
 
+variable "datawatch_rds_options" {
```
BrianKopp, line 1395: Do we also need this for the replica?

BrianKopp (approved):

modules/bigeye/variables.tf
deeno35: Yeah it's reasonable someone will want to set these differently, I'll add it.

PR #265, opened 2024-05-22
refactor: control opensearch redeundancy with var.redundant_infrastru…

deeno35:
…cture

There were several knobs controlling opensearch redunancy.  Hard code these in so it is all driven from the var.redundant_infrastructure varaible to streamline the interface and avoid mistakes.

BREAKING_CHANGE: The following variables have been removed
- temporal_opensearch_instance_count
- temporal_opensearch_enable_master_nodes
- temporal_opensearch_zone_awareness_zone_count

These are now no longer settable, but instead controlled via var.redundant_infrastructure.

### DIFF
--- a/modules/bigeye/variables.tf
+++ b/modules/bigeye/variables.tf
@@ -1002,12 +1002,6 @@ variable "temporal_opensearch_instance_type" {
   default     = "t3.medium.search"
 }
 
-variable "temporal_opensearch_instance_count" {
-  description = "The number of data nodes"
-  type        = number
-  default     = 1
-}
-
 variable "temporal_opensearch_master_user_password_secret_arn" {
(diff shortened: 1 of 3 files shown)


modules/opensearch/variables.tf
```diff
@@ -64,7 +64,7 @@ variable "master_node_instance_type" {
 }
 
 variable "zone_awareness_zone_count" {
```
BrianKopp, line 66: Could we just remove this variable since it's not being passed anymore?

modules/bigeye/variables.tf
BrianKopp, line 1: This will be a lot simpler to manage. Good change here

BrianKopp (approved):
Nice!

PR #268, opened 2024-05-31
feat: allow using firelens for ECS logs

deeno35:
Allow sending ECS logs via AWS firelens.  This opens the door for shipping logs to other log destinations besides Cloudwatch.

### DIFF
--- a/modules/bigeye/temporal.tf
+++ b/modules/bigeye/temporal.tf
@@ -126,7 +126,11 @@ resource "aws_ecs_task_definition" "temporal_components" {
   requires_compatibilities = ["FARGATE"]
   tags                     = merge(local.tags, { app = "temporal", component = each.key })
   execution_role_arn       = local.ecs_role_arn
-  container_definitions    = var.datadog_agent_enabled ? jsonencode([local.temporal_component_container_def[each.key], local.temporal_component_datadog_container_def[each.key]]) : jsonencode([local.temporal_component_container_def[each.key]])
+  container_definitions = jsonencode(concat(
+    [local.temporal_component_container_def[each.key]],
+    var.datadog_agent_enabled ? [local.temporal_component_datadog_container_def[each.key]] : [],
+    var.awsfirelens_enabled && var.temporal_logging_enabled ? [local.temporal_component_awsfirelens_container_def[each.key]] : [],
+  ))
 }
 
 resource "aws_ecs_service" "temporal_components" {
@@ -416,7 +420,17 @@ locals {
     var.temporal_additional_secret_arns,
   )
 
-  log_configuration_def = var.temporal_logging_enabled ? {
+  log_configuration_def = var.temporal_logging_enabled ? var.awsfirelens_enabled ? {
+    logDriver = "awsfirelens",
+    options = {
+      "Name"       = "http"
+      "Host"       = var.awsfirelens_host
+      "URI"        = var.awsfirelens_uri
+      "Port"       = 443
+      "tls"        = "on"
+      "tls.verify" = "off"
+      "format"     = "json_lines"
+    } } : {
     logDriver = "awslogs"
     options = {
       "awslogs-group"         = aws_cloudwatch_log_group.temporal.name
@@ -491,8 +505,8 @@ locals {
       local.temporal_container_def_general,
       {
         name   = "${local.name}-temporal-${local.temporal_svc_override_names[svc]}"
-        cpu    = var.datadog_agent_enabled ? local.temporal_component_cpu[svc] - var.datadog_agent_cpu : local.temporal_component_cpu[svc]
-        memory = var.datadog_agent_enabled ? local.temporal_component_memory[svc] - var.datadog_agent_memory : local.temporal_component_memory[svc]
+        cpu    = local.temporal_component_cpu[svc] - (var.datadog_agent_enabled ? var.datadog_agent_cpu : 0) - (var.awsfirelens_enabled ? var.awsfirelens_cpu : 0)
+        memory = local.temporal_component_memory[svc] - (var.datadog_agent_enabled ? var.datadog_agent_memory : 0) - (var.awsfirelens_enabled ? var.awsfirelens_memory : 0)
         dockerLabels = var.datadog_agent_enabled ? merge(
           local.temporal_docker_labels_general,
           {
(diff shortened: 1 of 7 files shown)


modules/bigeye/temporal.tf
```diff
@@ -416,7 +420,18 @@ locals {
     var.temporal_additional_secret_arns,
   )
 
-  log_configuration_def = var.temporal_logging_enabled ? {
+  log_configuration_def = var.temporal_logging_enabled ? var.awsfirelens_enabled ? {
```
deeno35, line 423: This is pretty ugly, but I don't know of another way to allow a "nested if" with two flags and effectively 3 "branches"

modules/bigeye/variables.tf
```diff
@@ -276,6 +276,45 @@ variable "datadog_agent_api_key_secret_arn" {
...
+#======================================================
+# AWS Firelens settings
+#======================================================
+variable "awsfirelens_enabled" {
+  description = "Whether to include the awsfirelens container in the task definition"
+  type        = bool
+  default     = false
+}
+
+variable "awsfirelens_image" {
```
BrianKopp, line 288: Is there a good reason to have this image be specified in a different way than our other images? Like could we have `_registry` that defaults to the overall `image_registry`, `_repository` that defaults (or doesn't), and then the `_image_tag`, that defaults to the `image_tag`? Or is this different enough that it's worth it being different

modules/bigeye/variables.tf
```diff
@@ -276,6 +276,45 @@ variable "datadog_agent_api_key_secret_arn" {
...
+  description = "The full image for aws firelens, e.g. registry-host-name.com/repository/name:tag.  It is recommended to pin this to a specific tag for production systems vs relying on latest."
+  type        = string
+  default     = "amazon/aws-for-fluent-bit:latest"
```
BrianKopp, line 291: Oh, this lol. This default is a probably a good enough reason to have it be different lol

modules/bigeye/variables.tf
```diff
@@ -731,13 +770,13 @@ variable "haproxy_desired_count" {
 variable "haproxy_cpu" {
   description = "Amount of CPU to allocate"
   type        = number
-  default     = 512
+  default     = 1024
```
deeno35, line 773: min container size and mem had to be bumped for containers that are smaller than the default cpu for datadog + awsfirelens (256+256=512).

modules/bigeye/variables.tf
```diff
@@ -276,6 +276,45 @@ variable "datadog_agent_api_key_secret_arn" {
...
+variable "awsfirelens_cpu" {
+  description = "The amount of CPU to allocate to the AWS firelens container, in Mhz, e.g. 256"
+  type        = number
+  default     = 256
+}
+
+variable "awsfirelens_memory" {
+  description = "The amount of Memory to allocate to the AWS firelens container, in MiB, e.g. 512"
+  type        = number
+  default     = 100
```
BrianKopp, line 303: Is 100 a valid memory option?

modules/bigeye/variables.tf
```diff
@@ -276,6 +276,45 @@ variable "datadog_agent_api_key_secret_arn" {
...
+}
+
+variable "awsfirelens_host" {
+  description = "Use this setting and awsfirelens_URI to send logs via awsfirelens instead of to cloudwatch logs.  Mutually exclusive with var.cloudwatch_log_group_name"
+  type        = string
+  default     = "endpoint4.collection.sumologic.com"
```
BrianKopp, line 309: Should we have this in as our default here?

modules/simpleservice/variables.tf
```diff
@@ -153,7 +153,7 @@ variable "docker_labels" {
 }
 
 variable "cloudwatch_log_group_name" {
-  description = "Cloudwatch log group name to write container logs"
+  description = "Cloudwatch log group name to write container logs.  If var.awsfirelens_enabled = true, container logs are shipped via AWS firelens instead.  The AWS firelens side car container will continue to send logs to cloudwatch logs even if var.awsfirelens_enabled = true to facilitate debugging firelens issues."
```
BrianKopp, line 156: Does this mean it's dual sending? I don't understand yet. maybe I will when I finish reading this PR

modules/simpleservice/variables.tf
```diff
@@ -230,6 +230,46 @@ variable "datadog_agent_additional_secret_arns" {
...
+variable "awsfirelens_image" {
+  description = "The full image for aws firelens, e.g. registry-host-name.com/repository/name:tag.  It is recommended to pin this to a specific tag for production systems vs relying on latest."
+  type        = string
+  default     = "amazon/aws-for-fluent-bit:latest"
+}
+
+variable "awsfirelens_cpu" {
+  description = "The amount of CPU to allocate to the AWS firelens container, in Mhz, e.g. 256"
+  type        = number
+  default     = 256
```
BrianKopp, line 252: Can we remove the defaults on all these since we'll be passing them from the bigeye module. It just adds a place we need to think about defaults. Maybe leave the defaults on the _enabled vars tho

modules/simpleservice/containers.tf
```diff
@@ -26,7 +26,18 @@ locals {
...
+      options = {
+        "Name"       = "http"
+        "Host"       = var.awsfirelens_host
+        "URI"        = var.awsfirelens_uri
+        "Port"       = 443
+        "tls"        = "on"
+        "tls.verify" = "off"
+        "format"     = "json_lines"
+      }
+      } : {
```
BrianKopp, line 40: Alignment on these curly brackets makes it look like there's a doubletap

modules/simpleservice/containers.tf
```diff
@@ -93,12 +104,38 @@ locals {
...
+      }
+    }
+    logConfiguration = {
+      "logDriver" : "awslogs",
+      "options" : {
+        "awslogs-group"         = var.cloudwatch_log_group_name
+        "awslogs-region"        = data.aws_region.current.name
+        "awslogs-stream-prefix" = var.name
+      }
+    }
```
BrianKopp, line 129: So is this the double-send part?

BrianKopp (approved):

modules/bigeye/variables.tf
deeno35: This is much better, I'll update

modules/bigeye/variables.tf
deeno35: It's actually the value from the aws docs example.  The problem is we won't know it's a problem until we get it onto our highest volume service.  I'm kind of loathe to give it any more than is necessary, but it probably isn't the end of the world to give it a little more, since the unused amount would float.  I'll udpate it to 256

modules/bigeye/variables.tf
deeno35: I guess we should hide it.  Good call.

modules/simpleservice/variables.tf
deeno35: I'll update the description.  If this is confusing to you, it's going to be really confusing for others.

modules/simpleservice/containers.tf
deeno35: The firelens container logs to cloudwatch, the app container logs to firelens.  This is so there is some debug ability if there is a prob with the firelens endpoint

modules/simpleservice/containers.tf
deeno35: I'll put these all onto a single line, it's an improvement, but after trying a few different ways, tf fmt really doesn't do the right thing with this

PR #273, opened 2024-06-13
feat: add ability to mount EFS volume in app containers

deeno35:
This is useful for debugging, like in the case where we would want to persist some generated data on instance stop like a heap dump.

A single EFS volume for the entire stack was chosen instead of volume per service as ECS containers can only allow 1 EFS volume mount, so with a single mount, all of the various container's EFS data can be read from the bigeye-admin container.

EFS access point per service gives us logical separation of data between the different services, so each service can only read/write its own data (except the admin container which can mount the root "/" so will see them all).

Also note it is not possible to disable AZ redundancy by using "one zone" mode to save on costs as we run into problems with EFS mount points not existing in the various AZ's since we don't allow constraining ECS tasks to a single AZ.

### DIFF
--- a/modules/bigeye/main.tf
+++ b/modules/bigeye/main.tf
@@ -589,8 +589,7 @@ resource "aws_iam_role_policy" "ecs_execution" {
           "ecr:BatchGetImage"
         ]
         Resource = "*"
-      },
-      {
+        }, {
         Sid    = "AllowCloudWatch"
         Effect = "Allow"
         Action = [
@@ -604,7 +603,6 @@ resource "aws_iam_role_policy" "ecs_execution" {
       }
     ]
   })
-
 }
 
 resource "aws_iam_role_policy" "ecs_secrets" {
@@ -659,6 +657,9 @@ module "bigeye_admin" {
   execution_role_arn        = local.ecs_role_arn
   task_iam_role_arn         = var.admin_container_ecs_task_role_arn
   fargate_version           = var.fargate_version
+  efs_volume_id             = local.efs_volume_enabled && var.enable_bigeye_admin_module ? aws_efs_file_system.this[0].id : ""
+  efs_access_point_id       = local.efs_volume_enabled && var.enable_bigeye_admin_module ? aws_efs_access_point.bigeye_admin[0].id : ""
+  efs_mount_point           = var.efs_mount_point
 
   stack_name = local.name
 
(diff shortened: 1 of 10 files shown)


modules/bigeye/variables.tf
```diff
@@ -243,6 +243,30 @@ variable "fargate_version" {
...
+  description = "A list of additional security group ids to put onto the EFS volume"
+  type        = list(string)
+  default     = []
+}
+
+variable "efs_redundancy_enabled" {
+  description = "Redundant AZ infra for EFS, at the cost of $$$.  Leave this true, the mounts will fail on AZs without a mount point."
+  type        = bool
+  default     = true
+}
```
BrianKopp, line 262: Should we not include this as an option?

modules/bigeye/variables.tf
```diff
@@ -243,6 +243,30 @@ variable "fargate_version" {
   default     = "1.4.0"
 }
 
+variable "efs_volume_enabled_services" {
+  description = "A shared EFS volume can be mounted on core service containers.  This can be useful for containers to persist a heap dump on OOM for example.  Service names: datawatch, datawork, metricwork, lineagework, etc"
+  type        = list(string)
+  default     = []
+}
```
BrianKopp, line 250: Is there a benefit to having a list of services here or just having a global enabled/disabled? Seems like a potential maintenance item/foot gun

modules/bigeye/main.tf
```diff
@@ -1756,6 +1772,32 @@ resource "aws_iam_role_policy" "datawatch_ecs_exec" {
   })
 }
 
+resource "aws_iam_role_policy" "datawatch_efs" {
+  count = local.create_datawatch_role && local.efs_volume_enabled && (var.datawatch_enable_ecs_exec || var.datawork_enable_ecs_exec || var.lineagework_enable_ecs_exec || var.metricwork_enable_ecs_exec) ? 1 : 0
```
BrianKopp, line 1776: Ok, so we'd only create this if it's in the list AND we have exec enabled?

BrianKopp: Should those choices be separated?

modules/bigeye/main.tf
```diff
@@ -1756,6 +1772,32 @@ resource "aws_iam_role_policy" "datawatch_ecs_exec" {
...
+  name  = "AllowECSExec"
+  policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [
+      {
+        "Effect" : "Allow",
+        "Action" : [
+          "elasticfilesystem:ClientMount",
+          "elasticfilesystem:ClientWrite"
+        ],
```
BrianKopp, line 1787: So our handful of services that have no IAM role just can't use this? E.g. haproxy/web?

BrianKopp (approved):

PR #287, opened 2024-06-25
feat: use KMS to encrypt secrets

szhongren:
create resources for KMS encryption

### DIFF
--- a/modules/bigeye/locals.tf
+++ b/modules/bigeye/locals.tf
@@ -176,6 +176,8 @@ locals {
       ROBOT_PASSWORD         = local.robot_password_secret_arn
       BASE_ENCRYPTION_SECRET = local.base_datawatch_encryption_secret_arn
       BASE_SALT              = local.base_datawatch_salt_secret_arn
+      KMS_KEY_ID             = aws_kms_key.datawatch.key_id
+      USE_KMS                = var.datwatch_encrypt_secrets_with_kms_enabled
     },
     var.datawatch_additional_secret_arns,
   )
--- a/modules/bigeye/main.tf
+++ b/modules/bigeye/main.tf
@@ -2029,6 +2029,16 @@ resource "aws_secretsmanager_secret_version" "base_salt" {
   secret_string  = random_password.base_salt[0].result
   version_stages = ["AWSCURRENT"]
 }
+resource "aws_kms_alias" "encryption_key_alias" {
+  name          = format("alias/bigeye/%s/datawatch", local.name)
+  target_key_id = aws_kms_key.datawatch.key_id
+}
+resource "aws_kms_key" "datawatch" {
+  description         = "KMS key that we use to encrypt/decrypt secrets. This will be used for securely storing sensitive information such as connection info. One will be created if not provided."
+  enable_key_rotation = true
+  # enable when we get past 5.33.0
+  # rotation_period_in_days = 120
+}
 
 locals {
   datawatch_common_env_vars = {
--- a/modules/bigeye/variables.tf
+++ b/modules/bigeye/variables.tf
@@ -1375,6 +1375,12 @@ variable "datawatch_base_salt_secret_arn" {
   default     = ""
 }
 
+variable "datwatch_encrypt_secrets_with_kms_enabled" {
+  description = "Flag for enabling KMS secret encryption. Setting this to false is not recommended unless KMS is not available in your AWS environment."
+  type        = bool
+  default     = true
+}
+
 variable "datawatch_db_name" {
   description = "The database name to use in the connection string for Datawatch apps. This is here because some RDS instances have been created with a blank DB name, so it can't be assumed the RDS DB name is a valid Database name for the connection string. If not specified, it will default to `datawatch_rds_db_name`"
   type        = string


modules/bigeye/variables.tf
```diff
@@ -1375,6 +1375,18 @@ variable "datawatch_base_salt_secret_arn" {
   default     = ""
 }
 
+variable "datawatch_use_kms" {
+  description = "Flag for turning on KMS secret encryption"
+  type        = bool
+  default     = true
+}
+
+variable "datawatch_base_encryption_key_id" {
```
deeno35, line 1384: If we don't want someone to be able to bring their own KMS key, we don't need this variable and just the boolean for flagging the feature on is enough

modules/bigeye/locals.tf
```diff
@@ -56,6 +56,8 @@ locals {
   datawatch_rds_password_secret_arn    = local.create_datawatch_rds_password_secret ? aws_secretsmanager_secret.datawatch_rds_password[0].arn : var.datawatch_rds_root_user_password_secret_arn
   create_temporal_rds_password_secret  = var.temporal_rds_root_user_password_secret_arn == ""
   temporal_rds_password_secret_arn     = local.create_temporal_rds_password_secret ? aws_secretsmanager_secret.temporal_rds_password[0].arn : var.temporal_rds_root_user_password_secret_arn
+  create_base_kms_key                  = var.datawatch_base_encryption_key_id == ""
```
deeno35, line 59: Are we ever going to allow someone to pass in their own kms key for this?  If not the flag should be named <feature>_enabled

IE encrypt_secrets_with_kms_enabled

modules/bigeye/locals.tf
```diff
@@ -56,6 +56,8 @@ locals {
...
+  base_datawatch_encryption_key_id     = local.create_base_kms_key ? aws_kms_key.encryption_key.key_id : var.datawatch_base_encryption_key_id
```
deeno35, line 60: This can just be the following if we don't allow someone to bring their own KMS key.

> local.create_base_kms_key ? aws_kms_key.encryption_key.key_id : ""

modules/bigeye/variables.tf
```diff
@@ -1375,6 +1375,18 @@ variable "datawatch_base_salt_secret_arn" {
   default     = ""
 }
 
+variable "datawatch_use_kms" {
```
deeno35, line 1378: Our feature flags are all named to end with _enabled.  We probably want to be specific about what this kms key is used for, as there could be other kms keys down the road.

Example name: encrypt_secrets_with_kms_enabled

We would want to be pretty verbose in the description to tell users what it means when this flag is set to false.  IE what is actually encrypted with KMS when this flag is enabled and how will it be secured if the flag is false.  For a flag like this, I would also included a sentence at the end like "Setting this to false is not recommended unless KMS is not available in your AWS environment."

modules/bigeye/locals.tf
```diff
@@ -176,6 +178,7 @@ locals {
       ROBOT_PASSWORD         = local.robot_password_secret_arn
       BASE_ENCRYPTION_SECRET = local.base_datawatch_encryption_secret_arn
       BASE_SALT              = local.base_datawatch_salt_secret_arn
+      KMS_KEY_ID             = local.base_datawatch_encryption_key_id
```
deeno35, line 181: nit: this could just be named datawatch_kms_key_id unless we plan on having more kms keys

modules/bigeye/main.tf
```diff
@@ -2029,6 +2029,16 @@ resource "aws_secretsmanager_secret_version" "base_salt" {
   secret_string  = random_password.base_salt[0].result
   version_stages = ["AWSCURRENT"]
 }
+resource "aws_kms_alias" "encryption_key_alias" {
+  name          = format("alias/bigeye/%s/datawatch/encryption-key", local.name)
```
deeno35, line 2033: nit: probably we can shorten this some since alias and encryption-key don't really add much.  

bigeye/%s/datawatch

modules/bigeye/main.tf
```diff
@@ -2029,6 +2029,16 @@ resource "aws_secretsmanager_secret_version" "base_salt" {
...
+  target_key_id = aws_kms_key.encryption_key.key_id
+}
+resource "aws_kms_key" "encryption_key" {
```
deeno35, line 2036: You can use datawatch instead of encryption_key as the resource name.  If we will ever have more than 1 encryption key for a different purpose, then

datawatch_<purpose> might be a good resource name.

modules/bigeye/main.tf
```diff
@@ -2029,6 +2029,16 @@ resource "aws_secretsmanager_secret_version" "base_salt" {
...
+  description         = "Auto rotated key used for encryption/decryption"
```
deeno35, line 2037: We cut paste in the description from the variable in here.  IE if someone was browsing AWs, they would see something like

Bigeye uses this key to xxxx. 

We don't need to mention the auto rotation, someone who is looking at the key will see how it is set.

modules/bigeye/variables.tf
szhongren: we need to pass the KMS key to config.yaml though, does that not need this too?

deeno35:
The git commit subject and body end up in our publicly accessible release notes.  A better subject body would describe the feature and what it does (think external comments, not internal).  ie

feat: use KMS to encrypt xxxxxx

Using KMS is more secure than __previous method described__.  It allows autorotation, blah.  This feature allows Bigeye to use KMS instead of custome blah

modules/bigeye/variables.tf
deeno35: You actually don't since the terraform is creating the KMS resource so you have it with:

KMS_KEY_ID             = aws_kms_key.encryption_key.key_id

modules/bigeye/variables.tf
szhongren: made this change

modules/bigeye/locals.tf
szhongren: changed

modules/bigeye/main.tf
szhongren: alias is required, but will remove encryption-key

modules/bigeye/main.tf
szhongren: done

deeno35 (approved):

PR #306, opened 2024-07-18
fix: rename papi service to internalapi

deeno35:
papi (private api) ended up not being a popular name internally, so it is being renamed to internalapi.  It was a thought to support the existing variable names and avoid the breaking change, but it will keep the house much cleaner if we just take the breaking change and remove them.

BREAKING CHANGE: All variables with papi in the name need to be globally replaced with internalwatch.

### DIFF
--- a/modules/admin/main.tf
+++ b/modules/admin/main.tf
@@ -29,7 +29,7 @@ locals {
     DATAWORK_DOMAIN_NAME    = var.datawork_domain_name
     LINEAGEWORK_DOMAIN_NAME = var.lineagework_domain_name
     METRICWORK_DOMAIN_NAME  = var.metricwork_domain_name
-    PAPI_DOMAIN_NAME        = var.papi_domain_name
+    INTERNALAPI_DOMAIN_NAME = var.internalapi_domain_name
     SCHEDULER_DOMAIN_NAME   = var.scheduler_domain_name
 
     HAPROXY_ELB_NAME     = var.haproxy_resource_name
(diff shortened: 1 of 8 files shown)


modules/admin/main.tf
```diff
@@ -29,7 +29,7 @@ locals {
     DATAWORK_DOMAIN_NAME    = var.datawork_domain_name
     LINEAGEWORK_DOMAIN_NAME = var.lineagework_domain_name
     METRICWORK_DOMAIN_NAME  = var.metricwork_domain_name
-    PAPI_DOMAIN_NAME        = var.papi_domain_name
+    PAPI_DOMAIN_NAME        = var.internalwatch_domain_name
```
BrianKopp, line 32: This env var name be changed as well. Same as others in the file.

modules/alarms/main.tf
```diff
@@ -676,16 +676,16 @@ module "ecs_monocle" {
   mem_threshold           = var.ecs_monocle_mem_threshold
 }
 
-module "ecs_papi" {
+module "ecs_internalwatch" {
   source                  = "./ecs"
   stack                   = var.stack
   app                     = "monocle"
```
BrianKopp, line 682: This feels like a copy pasta error? Monocle?

modules/bigeye/variables.tf
```diff
@@ -1935,67 +1935,67 @@ variable "metricwork_enable_ecs_exec" {
 #======================================================
 # Application Variables - Papi
```
BrianKopp, line 1936: Can you update the heading here? Probably worth a global search in the repo for papi

BrianKopp (approved):

PR #311, opened 2024-07-25
feat: add env vars to control job executor parallelism

deeno35:
The job runners can consume too much cpu/mem as the defaults are for 200 parallel threads.  These environment variables allow fine grained control over all of the Temporal based work queues.

I opted for a generic name on the TF variables as the queue membership can be fluid and would become breaking changes if we moved queue subscription around to different services

### DIFF
--- a/modules/bigeye/main.tf
+++ b/modules/bigeye/main.tf
@@ -2293,12 +2293,24 @@ module "datawork" {
     local.datawatch_dd_env_vars,
     local.datawatch_common_env_vars,
     {
-      APP                = "datawork"
-      DATAWATCH_ADDRESS  = "http://localhost:${var.datawork_port}"
-      WORKERS_ENABLED    = "true"
-      METRIC_RUN_WORKERS = "1"
-      EXCLUDE_QUEUES     = "trigger-batch-metric-run,source-lineage,metacenter-lineage"
-      HEAP_DUMP_PATH     = contains(var.efs_volume_enabled_services, "datawork") ? var.efs_mount_point : ""
+      APP                              = "datawork"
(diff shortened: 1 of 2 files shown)


modules/bigeye/main.tf
```diff
@@ -2374,13 +2386,17 @@ module "lineagework" {
...
-      WORKERS_ENABLED    = "true"
-      METRIC_RUN_WORKERS = "1"
-      EXCLUDE_QUEUES     = "run-metrics.v1,delete-source.v1,get-samples.v1,collect-lineage.v1,indexing.v1,reconciliation,trigger-batch-metric-run,agent-heartbeat"
-      MQ_WORKERS_ENABLED = "false"
-      HEAP_DUMP_PATH     = contains(var.efs_volume_enabled_services, "lineagework") ? var.efs_mount_point : ""
+      APP                          = "lineagework"
+      DATAWATCH_ADDRESS            = "http://localhost:${var.lineagework_port}"
+      WORKERS_ENABLED              = "true"
+      METRIC_RUN_WORKERS           = "1"
+      EXCLUDE_QUEUES               = "run-metrics.v1,delete-source.v1,get-samples.v1,collect-lineage.v1,indexing.v1,reconciliation,trigger-batch-metric-run,agent-heartbeat,refresh-scorecards"
```
deeno35, line 2393: One bug fix hiding in here.  The refresh-scorecards queue was added, but was not in the exclude queue for lineagework so I added it

deeno35:
Also 2 queues were omitted: agent_heartbeat and collect-lineage.v1.  Those are java agent temporal queues and not serviced by the Bigeye app stack itself.

BrianKopp (approved):

PR #315, opened 2024-07-28
Tune default reconciliation workflows pollers

jfcole:
See also https://github.com/bigeyedata/semantic-data-platform/pull/14630


deeno35 (approved):

deeno35:
We should probably run this for a while (set it per stack), before setting the default

jfcole:
I don't know how much tuning matters for this one before we make the change in the linked PR. There's only ever a single workflow triggered by this schedule, with a single activity. I'm hoping to get a sense for tuning in staging, and adjust accordingly before further rollout if necessary. 

I assume I'll have to make this change different to do it stack by stack?

PR #319, opened 2024-08-08
feat: new Temporal task queue monocle-invalidation

mhalverson:
As requested in bigeyedata/semantic-data-platform#14885

Fixes ONE-2704

### DIFF
--- a/modules/bigeye/main.tf
+++ b/modules/bigeye/main.tf
@@ -2293,24 +2293,26 @@ module "datawork" {
     local.datawatch_dd_env_vars,
     local.datawatch_common_env_vars,
     {
-      APP                              = "datawork"
-      DATAWATCH_ADDRESS                = "http://localhost:${var.datawork_port}"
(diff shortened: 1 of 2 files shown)


deeno35 (approved):

deeno35:
Looks like the linter,  if you run `terraform fmt` in the dir, it will autofix

deeno35:
This is cool, I am very excited about anything to do with the trained models going over to temporal

deeno35:
Also feel free to merge when green, there is still one more gate before it ends up on production, so Brian and I will check it out and make sure it's all good before rollout

mhalverson:
Thanks! This is pretty minimal; just making the HTTP call to Monocle's `/invalidate` endpoint occur asynchronously. Personally I am more excited by how easy it was to add a new Temporal workflow. I think we've hit the tipping point; it's easier to use Temporal than Rabbit!

jfcole (approved):

PR #323, opened 2024-08-20
fix: reduce default mc lineage workflow task pollers

jfcole:


deeno35 (approved):

deeno35:
In case you didn't know, you can run `terraform fmt` in the directory and it will reformat the file

PR #329, opened 2024-09-05
feat: allow setting JVM MaxRAMPercentage for each service

deeno35:
These datawatch dirivative services were all sharing the same variable, but each service operates differently so should have their own setting

### DIFF
--- a/modules/bigeye/main.tf
+++ b/modules/bigeye/main.tf
@@ -2120,7 +2120,6 @@ locals {
     FF_ANALYTICS_LOGGING_ENABLED    = var.datawatch_feature_analytics_logging_enabled
     FF_QUEUE_BACKFILL_ENABLED       = "true"
     FF_SEND_ANALYTICS_ENABLED       = var.datawatch_feature_analytics_send_enabled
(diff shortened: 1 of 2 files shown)


modules/bigeye/main.tf
```diff
@@ -2293,6 +2293,7 @@ module "datawork" {
       APP                                = "datawork"
       DATAWATCH_ADDRESS                  = "http://localhost:${var.datawork_port}"
       WORKERS_ENABLED                    = "true"
+      MAX_RAM_PERCENTAGE                 = var.datawork_jvm_max_ram_pct
```
BrianKopp, line 2296: Do we need to consider this a breaking change? Someone with `datawatch_jvm_max_ram_pct` set to non-default will see different settings to other derivative services?

BrianKopp (approved):

modules/bigeye/main.tf
deeno35: The behavior will change, but I don't think it's a breaking change.  I had thought about using a single setting with overrides for each service, but really the services are all independent as their memory profiles and behavior are all different.  So just having simple knobs for each service seems to feel good here (KISS)

PR #340, opened 2024-10-11
feat: use an include queue list instead of exclude queue

deeno35:
There is a new feature in datawatch and related services to allow an include list.  Using this removes the maintenance task of needing to maintain the exclude list everytime a new queue is added to temporal.

Also convert metricwork over to use this instead of the single queue override.  That feature is (or should be) deprecated and removed since we now have an include list.

This shouldn't be merged until 1.71.0 is released.

### DIFF
--- a/modules/bigeye/main.tf
+++ b/modules/bigeye/main.tf
@@ -2382,7 +2382,6 @@ module "lineagework" {
   awsfirelens_image   = var.awsfirelens_image
   awsfirelens_uri     = var.awsfirelens_uri
 
-  # lineagework should only subscribe to source-lineage and metacenter-lineage, ie exclude all other Temporal queues.
   environment_variables = merge(
     local.datawatch_dd_env_vars,
     local.datawatch_common_env_vars,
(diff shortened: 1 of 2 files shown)


modules/bigeye/main.tf
```diff
@@ -2392,7 +2391,7 @@ module "lineagework" {
       WORKERS_ENABLED              = "true"
       MAX_RAM_PERCENTAGE           = var.lineagework_jvm_max_ram_pct
       METRIC_RUN_WORKERS           = "1"
-      EXCLUDE_QUEUES               = "run-metrics.v1,delete-source.v1,get-samples.v1,collect-lineage.v1,indexing.v1,reconciliation,trigger-batch-metric-run,agent-heartbeat,refresh-scorecards,monocle-invalidation,issue-root-cause"
+      INCLUDE_QUEUES               = "source-lineage,metacenter-lineage"
```
mhalverson, line 2394: I compared against `TaskQueue`. This will start excluding `issue-update` and `external-ticket`. I think it was a bug that they were not excluded, so good job for moving to a much safer pattern.

mhalverson (approved):

modules/bigeye/main.tf
deeno35: Yeah part of the motivation to clean house was seeing those dangling queues unexpectedly running on lineagework.

PR #343, opened 2024-10-25
fix: use a set instead of count in redis and RDS security groups 1/2

deeno35:
This is a bug.  Using the count will result in conflicts if the list membership ever change.  The canonical TF way is to use a for_each on a set as this prevents it.

We will need this in a subsequent PR when we add the Indexwork service


deeno35:
This just won't work because the inputs are dynamic so we have to use a "count".

PR #347, opened 2024-10-31
feat: autoscale indexwork service

deeno35:
This is an MQ based service so lends itself naturally to autoscaling.

Catalog indexing is not a real-time operation so autoscaling down to 0 when the queue is empty is ideal

### DIFF
--- a/modules/bigeye/main.tf
+++ b/modules/bigeye/main.tf
@@ -2346,7 +2346,7 @@ module "datawork" {
 }
 
 locals {
-  # TODO clean this up when indexwork_enabled featureflag is cleaned up SRE-3855
+  # TODO clean this up when indexwork_enabled featureflag is cleaned up SRE-3866
   indexwork_autoscaling_enabled = var.indexwork_enabled && var.indexwork_autoscaling_enabled
 }
 
@@ -2438,7 +2438,7 @@ resource "aws_appautoscaling_target" "indexwork" {
   count              = local.indexwork_autoscaling_enabled ? 1 : 0
   depends_on         = [module.indexwork]
   min_capacity       = 0
-  max_capacity       = 100
+  max_capacity       = var.indexwork_autoscaling_max_count
   resource_id        = format("service/%s/%s-indexwork", local.name, local.name)
   scalable_dimension = "ecs:service:DesiredCount"
   service_namespace  = "ecs"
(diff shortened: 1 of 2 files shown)


modules/bigeye/main.tf
```diff
@@ -2428,6 +2434,74 @@ module "indexwork" {
   secret_arns = local.datawatch_secret_arns
 }
 
+resource "aws_appautoscaling_target" "indexwork" {
+  count              = local.indexwork_autoscaling_enabled ? 1 : 0
+  depends_on         = [module.indexwork]
+  min_capacity       = 0
```
deeno35, line 2440: autoscale to 0 when the queue is empty

modules/bigeye/variables.tf
```diff
@@ -2044,6 +2044,49 @@ variable "indexwork_enable_ecs_exec" {
...
+variable "indexwork_autoscaling_threshold_step1" {
+  description = "The first autothreshold metric step"
+  type        = number
+  default     = 1
+}
+
+variable "indexwork_autoscaling_threshold_step2" {
+  description = "The second autothreshold metric step"
+  type        = number
+  default     = 1000
```
deeno35, line 2062: It's around 1.5s per message on adhoc01 to process indexing work so this is around 11 minutes of work so should be fine for a small stack (ie sane default).

modules/bigeye/variables.tf
```diff
@@ -2044,6 +2044,49 @@ variable "indexwork_enable_ecs_exec" {
...
+}
+
+variable "indexwork_autoscaling_threshold_step3" {
+  description = "The third and final autothreshold metric step"
+  type        = number
+  default     = 10000
```
deeno35, line 2068: At 1.5s per message, it would take 2 hosts 1hr to process 10k messages.  Seems like a more aggressive default than necessary, but since we autoscale to 0, there isn't a huge drawback to processing the work a list faster as long as the data source can handle the parallel queries.

modules/bigeye/main.tf
esmioley: What is the rate of scaledown? I'm mostly wondering since we ack immediately if we can run into issues where the work is acked, we scale down and kill the running job, and then we're not doing the work at all?

deeno35: Yeah this is a concern since our MQ workers don't respect a sigterm and continue to pluck work.  The scale will only change every 10 minutes (there is a cooldown setting set to 600), then the host waits 2minutes before it stops, but in the end we will still likely lose some work every time.

As a stop gap, maybe we should just scale in 2 steps instead of 4.  Basically scale to 0 when there are no jobs.  Scale to "desired count" if there is >= 1.  That should avoid mid-job task termination

esmioley: Based on offline conversation, these tasks are small enough that single-step reduction to 0 should not manage to cause issues with what's currently on the queue; the operations take <5s to happen.

esmioley (approved):

PR #370, opened 2024-11-12
fix: update autoscaling for services scaling on multiple queues

deeno35:
This method of using several autoscaling policies just leads to the two fighting for control over the number of instances.  Just use a cloudwatch metric that sums the message counts into a single metric instead

### DIFF
--- a/modules/bigeye/main.tf
+++ b/modules/bigeye/main.tf
@@ -2454,6 +2454,98 @@ module "backfillwork" {
   secret_arns = local.datawatch_secret_arns
 }
 
+resource "aws_appautoscaling_target" "backfillwork" {
+  depends_on         = [module.backfillwork]
+  min_capacity       = 0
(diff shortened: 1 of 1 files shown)


modules/bigeye/main.tf
```diff
@@ -2454,6 +2454,98 @@ module "backfillwork" {
...
+  alarm_actions       = [aws_appautoscaling_policy.backfillwork.arn]
+  evaluation_periods  = 1
+  datapoints_to_alarm = 1
+  threshold           = 0
+  comparison_operator = "GreaterThanOrEqualToThreshold"
+  treat_missing_data  = "missing"
+  tags                = {}
+  # (12 unchanged attributes hidden)
+
+  metric_query {
```
deeno35, line 2506: Use a single metric query that sums the message counts on all relevant queues instead of using multiple autoscaling policies

esmioley (approved):

PR #372, opened 2024-11-12
feat!: roll out lineage and catalog indexing queue migrations

deeno35:
This change has been vetted and is ready for 100% rollout so the feature flags are being removed.

The following queues will migrate off of datawork:
- lineage MQ queue will be serviced from lineagework
- catalog_index_v2 MQ queue will be serviced from indexwork

BREAKING CHANGE:

The following bigeye module variables have been removed:
- migrate_lineage_mq_queue_enabled
- migrate_catalog_indexing_mq_queue_enabled

Remove these from your configuration before upgrading to this version of the Bigeye terraform module.  If these have not been set before, no action is required.

### DIFF
--- a/modules/bigeye/locals.tf
+++ b/modules/bigeye/locals.tf
@@ -43,24 +43,27 @@ locals {
 
   datawork_mq_exclude_queues = join(",",
     concat(
-      compact([
-        "dataset_index_op_v2",
-        "metric_batch",
-        var.migrate_lineage_mq_queue_enabled ? "lineage" : "",
-        var.migrate_catalog_indexing_mq_queue_enabled ? "catalog_index_v2" : "",
-      ]),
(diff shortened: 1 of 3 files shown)


modules/bigeye/locals.tf
```diff
@@ -43,24 +43,27 @@ locals {
...
-      ]),
       local.backfillwork_mq_include_queues,
+      local.indexwork_mq_include_queues,
+      local.lineagework_mq_include_queues,
+      local.metricwork_mq_include_queues,
     )
   )
 
-  indexwork_mq_include_queues = join(",", compact([
+  indexwork_mq_include_queues = compact([
```
deeno35, line 53: I updated these all to look the same now where the var_str is the one that is used in the services.

roman-bigeye (approved):

PR #380, opened 2024-12-05
allow alb reqest count per target autoscaling for internalapi

deeno35:
commit fd178cc7103bbfb08e60574e249b39a6b672cebf (HEAD -> david/sre-3975-increase-the-base-instance-count-for-internalapi-to-more, origin/david/sre-3975-increase-the-base-instance-count-for-internalapi-to-more)
Date:   Wed Dec 4 15:49:38 2024 -0800

    feat!: normalize target autoscaling variable names

    monocle is also a target scaling service.  Make the variables look
    the same as internalapi.

    BREAKING CHANGE:

    The following vars have been removed:
    - var.monocle_autoscaling_enabled
    - var.monocle_max_count
    - var.monocle_autoscaling_request_count_target
    - var.indexwork_autoscaling_enabled

    The following vars have been added:
    - var.monocle_autoscaling_min_capacity
    - var.monocle_autoscaling_max_capacity
    - var.monocle_autoscaling_type
    - var.monocle_autoscaling_target_utilization

commit 48a9d30e8c3121451e9189c2cad9a370f096717c
Date:   Wed Dec 4 14:50:08 2024 -0800

    feat!: allow alb reqest count per target autoscaling for internalapi

    Allow control over the autoscaling algorithm as request count per target
    is more sensitive, but requires obersvation and babysitting.  If CPU
    scaling is leading to OOM or other overloading due to bursts, switch to
    request count per target or increase the min capacity.

    BREAKING CHANGE:

    The following vars have been removed:
    - var.internalapi_autoscaling_cpu_enabled
    - var.internalapi_autoscaling_cpu_target

    The following vars have been added:
    - var.internalapi_autoscaling_type
    - var.internalapi_autoscaling_min_capacity
    - var.internalapi_autoscaling_max_capacity
    - var.internalapi_autoscaling_target_utilization

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,22 @@ terraform version and application version.
 
 ## Upgrading
 
+### Upgrading to 15.0.0
+
+The following vars have been removed:
+
+- var.monocle_autoscaling_enabled
+- var.monocle_max_count
+- var.monocle_autoscaling_request_count_target
+- var.indexwork_autoscaling_enabled
+- var.internalapi_autoscaling_cpu_enabled
+- var.internalapi_autoscaling_cpu_target
+
+Instead, use:
+
+- var.internalapi_autoscaling_config
+- var.monocle_autoscaling_config
+
 ### Upgrading to 14.0.0
 
 The following variable will need to be removed from your config if you
--- a/modules/bigeye/main.tf
+++ b/modules/bigeye/main.tf
@@ -1308,7 +1308,7 @@ module "monocle" {
   lb_access_logs_bucket_prefix = format("%s-%s", local.elb_access_logs_prefix, "monocle")
 
   # Task settings
-  control_desired_count     = var.monocle_autoscaling_enabled ? false : true
+  control_desired_count     = var.monocle_autoscaling_config.type == "none"
   desired_count             = var.monocle_desired_count
   cpu                       = var.monocle_cpu
   memory                    = var.monocle_memory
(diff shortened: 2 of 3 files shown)


modules/bigeye/variables.tf
```diff
@@ -572,9 +572,21 @@ variable "monocle_image_tag" {
 }
 
 variable "monocle_desired_count" {
-  description = "The desired number of replicas"
+  description = "Initial capacity before autoscaling adjustments"
```
deeno35, line 575: The monocle changes are probably a surprise as we really only need to change internalapi.  But basically we have 2 services that are using target autoscaling and if I was going to put in a breaking change for one, we might as well make the variables the same for both of them.

We do have 2 services using step autoscaling, but it has different inputs for the steps, so no point in trying to make those use the same vars.

modules/bigeye/variables.tf
```diff
@@ -572,9 +572,21 @@ variable "monocle_image_tag" {
...
   type        = number
-  default     = 1
+  default     = 2
```
deeno35, line 577: Monocle is also an API server.  Its minimum count should be 2 as well for redundancy.

modules/bigeye/variables.tf
```diff
@@ -572,9 +572,21 @@ variable "monocle_image_tag" {
...
+variable "monocle_autoscaling_min_capacity" {
+  description = "Autoscaling min capacity when autoscaling is enabled"
+  type        = number
+  default     = 2
+}
+
+variable "monocle_autoscaling_max_capacity" {
+  description = "Autoscaling max capacity when autoscaling is enabled"
+  type        = number
+  default     = 3
```
deeno35, line 589: We should probably have a larger max here since this is autoscaling, but 3 is the current number.  I'll update this to 6 in its own PR.

modules/bigeye/variables.tf
```diff
@@ -619,20 +631,18 @@ variable "monocle_lb_extra_security_group_ids" {
...
-variable "monocle_max_count" {
-  description = "The maximum number of monocle instances allowed"
-  type        = number
-  default     = 3
+variable "monocle_autoscaling_type" {
+  description = "Autoscaling metric type.  None = no autoscaling, cpu_utilization = target tracking using avg cpu utilization, request_count_per_target = target tracking using ALB request count per target"
+  type        = string
+  default     = "request_count_per_target"
+  validation {
+    condition     = contains(["none", "cpu_utilization", "request_count_per_target"], var.monocle_autoscaling_type)
```
deeno35, line 639: This part is worth reviewing carefully.  So this string will control the type of target autoscaling (or "none" to disable).  It won't cover step scaling cases, but even if we update the vars down the road to handle step autoscaling, I think this set of variables will still be usable for that too as they are generic.

modules/bigeye/variables.tf
```diff
@@ -619,20 +631,18 @@ variable "monocle_lb_extra_security_group_ids" {
...
+    error_message = "Must be one of: none, cpu_utilization, request_count_per_target"
+  }
 }
 
-variable "monocle_autoscaling_request_count_target" {
-  description = "The scaling target for number of requests per instance"
+variable "monocle_autoscaling_target_utilization" {
```
deeno35, line 644: The naming here matches AWS terminology for these settings (target utilization)

modules/bigeye/main.tf
```diff
@@ -1379,25 +1379,41 @@ module "monocle" {
 }
 
 resource "aws_appautoscaling_target" "monocle" {
-  count              = var.monocle_autoscaling_enabled ? 1 : 0
+  count              = var.monocle_autoscaling_type == "none" ? 0 : 1
```
deeno35, line 1382: These should all get wrapped into a target scaling module, but I'm going to leave this in place for now to make fewer changes and get this deployed.

modules/bigeye/variables.tf
```diff
@@ -572,9 +572,21 @@ variable "monocle_image_tag" {
...
 
 variable "monocle_desired_count" {
-  description = "The desired number of replicas"
+  description = "Initial capacity before autoscaling adjustments"
   type        = number
-  default     = 1
+  default     = 2
+}
+
+variable "monocle_autoscaling_min_capacity" {
```
deeno35, line 580: capacity (vs count) matches aws terminology.  I left desired_count alone as it is named this way across all services and would break a bunch of things just to be consistent.

modules/bigeye/main.tf
```diff
@@ -1379,25 +1379,41 @@ module "monocle" {
...
+    predefined_metric_specification {
+      predefined_metric_type = "ECSServiceAverageCPUUtilization"
+    }
+    target_value = var.monocle_autoscaling_target_utilization
+  }
+}
+
+resource "aws_appautoscaling_policy" "monocle_request_count_per_target" {
+  count              = var.monocle_autoscaling_type == "request_count_per_target" ? 1 : 0
+  depends_on         = [aws_appautoscaling_target.monocle]
```
roman-bigeye, line 1409: depends_on is excessive here because you are referencing this resource later in the config

README.md
```diff
@@ -49,6 +49,28 @@ terraform version and application version.
...
+Instead, use:
+
+- var.monocle_autoscaling_min_capacity
+- var.monocle_autoscaling_max_capacity
+- var.monocle_autoscaling_type
+- var.monocle_autoscaling_target_utilization
+- var.internalapi_autoscaling_type
+- var.internalapi_autoscaling_min_capacity
+- var.internalapi_autoscaling_max_capacity
+- var.internalapi_autoscaling_target_utilization
```
roman-bigeye, line 72: I think it would be cleaner to use one map variable to configure autoscaling for each service.
Something like this:
```
variable "monocle_autoscaling_config" {
  type = any
  default = {
    min_capacity       = 2
    max_capacity       = 3
    type               = "request_count_per_target"
    target_utilization = 15
  }
}
```
This would make it flexible for other scaling types and to disable it would be an empty map.`monocle_autoscaling_config = {}`.
It would also group all related vars in one place.

roman-bigeye (approved):

PR #396, opened 2025-01-16
feat: enable auto minor version upgrade for rabbitmq

deeno35:
Starting with 3.13, AWS requires auto minor version upgrade to be enabled.  This also has to be set _before_ trying to upgrade to 3.13.  So it is just best to set it to true now.

If someone has not touched the default already, leave it set to false.  It will get turned on when they upgrade to 3.12.x anyways and that is required before they upgrade to 3.13 as it is not possible to upgrade from 3.11.x -> 3.13 directly.

image

https://docs.aws.amazon.com/amazon-mq/latest/developer-guide/rabbitmq-version-management.html

### DIFF
--- a/modules/rabbitmq/main.tf
+++ b/modules/rabbitmq/main.tf
@@ -9,8 +9,11 @@ terraform {
 }
 
 resource "aws_mq_broker" "queue" {
-  broker_name                = var.name
-  auto_minor_version_upgrade = false
+  broker_name = var.name
+  # If someone has changed from the original 3.11.20, enable auto minor version upgrade as that is
+  # what AWS requires starting with 3.13 anyways.
+  auto_minor_version_upgrade = var.engine_version == "3.11.20" ? false : true
+  apply_immediately          = true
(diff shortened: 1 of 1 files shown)


modules/rabbitmq/main.tf
```diff
@@ -9,8 +9,11 @@ terraform {
 }
 
 resource "aws_mq_broker" "queue" {
-  broker_name                = var.name
-  auto_minor_version_upgrade = false
+  broker_name = var.name
+  # If someone has changed from the original 3.11.20, enable auto minor version upgrade as that is
+  # what AWS requires starting with 3.13 anyways.
+  auto_minor_version_upgrade = var.engine_version == "3.11.20" ? false : true
+  apply_immediately          = true
```
deeno35, line 16: I snuck apply_immediately in here.  We generally want changes to go out when they are applied, vs at an off hour when staff may not be around to respond.

roman-bigeye (approved):

PR #400, opened 2025-01-17
feat: enable datawatch_request_body_logging_enabled everywhere

deeno35:
This flag just controls weather a trace id is logged in the response log from the backend server.  It will be set on (flag control removed) in Bigeye app version 2.2.0 anyways, so we are enabling this across the board in the terraform ahead of the code being released.

This variable will be removed in a future release.

### DIFF
--- a/modules/bigeye/variables.tf
+++ b/modules/bigeye/variables.tf
@@ -1951,9 +1951,9 @@ variable "datawatch_feature_jvm_memory_leak_fix" {
 }
 
 variable "datawatch_request_body_logging_enabled" {
-  description = "Whether request body logs are enabled"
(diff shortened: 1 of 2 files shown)


modules/bigeye/variables.tf
```diff
@@ -1951,9 +1951,9 @@ variable "datawatch_feature_jvm_memory_leak_fix" {
 }
 
 variable "datawatch_request_body_logging_enabled" {
-  description = "Whether request body logs are enabled"
+  description = "DEPRECATED: This flag will no longer do anything.  It will be removed in a future release"
```
deeno35, line 1954: Not deleting the flag now is just so we can group these minor things up vs having a major rev bump just for this

roman-bigeye (approved):

PR #410, opened 2025-01-28
feat: add tf module for Apache solr on ECS

roman-bigeye:


modules/solr-single-instance/runonce.sh
```diff
@@ -0,0 +1,129 @@
...
+REGION=$(get_metadata_and_curl http://169.254.169.254/latest/meta-data/placement/availability-zone | sed 's/.$//')
+export AWS_REGION=$REGION
+export AWS_DEFAULT_REGION=$REGION # for aws-cli v1 compatibility
+
+attach_volume() {
+    echo "Attaching volume $VAR_VOLUME_ID to instance $INSTANCE_ID as $DEVICE_NAME..."
+
+    aws ec2 attach-volume --volume-id "$VAR_VOLUME_ID" --instance-id "$INSTANCE_ID" --device "$DEVICE_NAME"
+
+    if [ $? -eq 0 ]; then
```
deeno35: Seems redundant to do all of this and have a while loop on the outside.  It would be a lot cleaner if there is just one call to attach_volume() and all of the logic surrounding retries etc is self contained.  So in the main area it becomes

```
if ! attach_volume(); then 
  error_message_and_exit
fi
```

Maybe even passing in the retry count and sleep interval as well

modules/solr-single-instance/runonce.sh
```diff
@@ -0,0 +1,129 @@
...
+        echo "Failed to attach volume $VAR_VOLUME_ID."
+        return 1  # Failure
+    fi
+}
+
+START_TIME=$SECONDS
+MAX_DURATION=180  # Total retry duration in seconds (3 minutes)
+while [ $((SECONDS - START_TIME)) -lt $MAX_DURATION ] && [ ! -b "$DEVICE_NAME" ]; do
+    attach_volume && break
+    echo "Retrying... (elapsed time: $((SECONDS - START_TIME)) seconds)"
```
deeno35, line 81: Also printing attempts remaining for wait loops is pretty nice, but I guess we aren't watching this log in realtime (only seen after everything finishes) so perhaps not useful

modules/solr-single-instance/variables.tf
```diff
@@ -0,0 +1,93 @@
+variable "solr_subnet" {
```
deeno35, line 1: would be good to follow the subnet naming for vars that were used in simpleservice

```
~/toro/terraform-modules/modules/simpleservice$ grep variable * | grep subnet
variables.tf:variable "subnet_ids" {
variables.tf:variable "lb_subnet_ids" {
```

modules/solr-single-instance/variables.tf
```diff
@@ -0,0 +1,93 @@
...
+  description = "VPC subnet where this instance will run."
+  type        = string
+}
+
+variable "alb_subnets" {
+  description = "VPC subnets for ALB attachments."
+  type        = list(string)
+}
+
+variable "env_instance_name" {
```
deeno35, line 11: this is just called "instance" in simpleservice, prob good to follow

modules/solr-single-instance/variables.tf
```diff
@@ -0,0 +1,93 @@
...
+  description = "Name of the environment instance."
+  type        = string
+}
+
+variable "service_name" {
```
deeno35, line 16: This is called "name" in simpleservice, probably good to follow

modules/solr-single-instance/variables.tf
```diff
@@ -0,0 +1,93 @@
...
+  type        = string
+}
+
+variable "vpc_id" {
+  description = "VPC ID where security group will be created."
+  type        = string
+}
+
+variable "solr_clients_sgs" {
+  description = "List of security group IDs that will have access to this solr server."
```
deeno35, line 27: I feel like we probably set something up for haproxy or temporal for ingress.  It would be good to handle this the same way so should not need a different pattern/variable naming convention here

modules/solr-single-instance/variables.tf
```diff
@@ -0,0 +1,93 @@
...
+  type        = list(string)
+  default     = []
+}
+
+variable "refresh_instance_on_launch_template_change" {
+  description = <<EOD
```
deeno35, line 33: Do we ever want to allow this to true?  I think no and we can just hard code it vs using a var?

modules/solr-single-instance/variables.tf
```diff
@@ -0,0 +1,93 @@
...
+    Whether or not the instance should be recreated immediately on every template change.
+    If false, existing instance will continue running but new instances will use the new template.
+    Note: changes happen when AWS releases new AMI, so we don't have control over that.
+  EOD
+  type        = bool
+  default     = false
+}
+
+variable "ecs_cluster_name" {
```
deeno35, line 42: I think this is var.ecs_cluster_id in simpleservice.  Would be good to follow

modules/solr-single-instance/variables.tf
```diff
@@ -0,0 +1,93 @@
...
+  type = string
+}
+
+variable "availability_zone" {
```
deeno35, line 46: Should we make this a forward looking variable and accept a list of availablilty zones and just create the single node instance in the first one vs having the caller manage it?

modules/solr-single-instance/variables.tf
```diff
@@ -0,0 +1,93 @@
...
+}
+
+variable "acm_certificate_arn" {
+  description = "ARN pointing to the certificate to terminate HTTPS traffic"
+  type        = string
+}
+
+variable "solr_traffic_port" {
+  type    = number
+  default = 8983
```
deeno35, line 69: I would like to normalize this port and use port 80/443.  I guess in this case 80 for running of the service itself since it does not comingle with any other processes.  It helps with audits if there are no strange ports to explain/review.

modules/solr-single-instance/variables.tf
```diff
@@ -0,0 +1,93 @@
...
+}
+
+variable "route53_zone_id" {
```
deeno35, line 72: Looks like our pattern for all of this has been to leave the DNS assignment at the top level module.  Do we need this inside the solr module because of service discovery?  If so that's fine, otherwise best to stick them together with the others vs passing through and doing it in the module

modules/solr-single-instance/variables.tf
```diff
@@ -0,0 +1,93 @@
...
+  default     = ""
+}
+
+variable "dns_name" {
+  description = "The name for Route53 DNS record."
+  type        = string
+  default     = ""
+}
+
+variable "elb_access_logs_bucket_config" {
```
deeno35, line 84: These are named this way for simpleservice, would be good to follow

```
var.lb_access_logs_enabled
var.lb_access_logs_bucket_name
var.lb_access_logs_bucket_prefix
```

If we like the map better, maybe we come back as a followup PR and fix simpleservice to match so there is consistency.

deeno35 (approved):

modules/bigeye/main.tf
```diff
@@ -2966,3 +2966,41 @@ resource "aws_appautoscaling_policy" "internalapi_request_count_per_target" {
     target_value = var.internalapi_autoscaling_config.target_utilization
   }
 }
+
+module "solr" {
+  count             = var.deploy_solr ? 1 : 0
+  source            = "../solr-single-instance"
```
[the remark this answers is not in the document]
roman-bigeye: That was the idea initially, but it is difficult to implement since we need to mount an EBS volume inside an instance. I think it will be easier to write a separate module if / when we decide to implement a clustered solution. It probably will not use ECS due to the lack of support for this use case.

modules/solr-single-instance/variables.tf
roman-bigeye: the idea behind taking a specific zone was to make sure that instances are distributed across multiple AZs if we have multiple of these deployed. This also highlights the fact that this is only one instance, so taking a list is king of pointless. I can change it to a list if you think it's better.

modules/solr-single-instance/runonce.sh
roman-bigeye: While loop is needed because previous instance shuts down and still have volume mounted when new is booting up, so it takes multiple retries until the volume is released.

modules/solr-single-instance/variables.tf
roman-bigeye: It is useful for testing and development. Also can be used to refresh instance if needed by code change instead of doing it via console manually.

modules/solr-single-instance/variables.tf
roman-bigeye: `cluster_name`  is the only input data. `aws_ecs_cluster` takes. `cluster_id` is ARN which is not suitable for this.
I use both in this module, so will have to either pass 2 vars or only name and find arn using data source.

modules/solr-single-instance/variables.tf
roman-bigeye: This port is used internally and is a standard port for this service. For external access it is behind ALB port 443.

modules/solr-single-instance/variables.tf
roman-bigeye: This zone_id is only used to create route53 record.
I think managing related resources in multiple places is a bad pattern. Why do we want to manage DNS records in the parent module?

I had to create`aws_ecs_cluster_capacity_providers` in parent because it can only be done by one resource and it makes sense to manage it next to the cluster in case we add more capacity providers in the future.

But I'd like to keep related resources contained in the same module if possible.
This makes adding more instances of this module and cleanup easier.

modules/solr-single-instance/variables.tf
deeno35: I am recommending we run every HTTP service internally on port 80, ie the service itself serves on 80.  The LB exposing SSL on 443 is fine.

We have updated all of the Bigeye services to port 80 already even though you'll notice the dev envs don't run on that port.  Basically this was a suggestion from a few of our customers to reduce the number of strange ports that need to be reviewed.  The upside for us is we don't have to maintain lists of strange port numbers for every service, it's just expected that they are running on 80 in AWS

modules/solr-single-instance/variables.tf
deeno35: It's more like if we end up with cluster down the road, the interface just works and would not need to change.  Possibly over optimizing.  For distrubution one could just modulo over the list too, but taking a single AZ then puts the burden of on the caller to implement.

If you feel strongly about it, feel free to leave it.  It's not an important thing and can be changed later, especially since it's our internal interface/var names and not the bigeye module's top level ones.

modules/solr-single-instance/variables.tf
deeno35: I don't disagree, but if you do this, now we end up with inconsistency.  So I would say if you like this and we follow up later and fixing them all so there is 1 way things are done, that's great.

modules/solr-single-instance/variables.tf
roman-bigeye: changed to 3 vars

modules/solr-single-instance/variables.tf
roman-bigeye: I'll follow up to update existing resources to follow this pattern as well.

modules/solr-single-instance/variables.tf
roman-bigeye: changed port to 80

modules/solr-single-instance/runonce.sh
roman-bigeye: moved retry logic inside attach_volume function

modules/solr-single-instance/runonce.sh
roman-bigeye: per our discussion we will keep this output as it can be useful for debugging.

PR #412, opened 2025-02-05
feat: configure solr image via variable

roman-bigeye:
This change switches bind mounts to generic paths and adds a variable for solr image.

This will allow us baking the static configs in the image to make sure that they are version controlled.
Now only data on the mounted EBS volume is the data modifiable by solr.


modules/bigeye/lineageplus.tf
```diff
@@ -20,4 +20,6 @@ module "lineageplus_solr" {
 
   service_discovery_private_dns_namespace_id = aws_service_discovery_private_dns_namespace.this.id
   refresh_instance_on_launch_template_change = true
+
+  solr_image = var.lineageplus_solr_image
```
deeno35, line 24: Looks like we are using "image_tag" as the naming convention for the others so solr_image_tag would make sense here (or lineageplus_solr_image_tag).

It works in conjunction with var.image_registry

deeno35 (approved):

modules/bigeye/lineageplus.tf
roman-bigeye: fixed

PR #413, opened 2025-02-07
feat: add option to use Cloudfront in front of the app

deeno35:
Using a CDN is a standard technique to improve UI performance for requests from latency distant regions.

The caller will pass in both the S3 logging bucket and ACM cert to use with Cloudfront as they are likely to already exist for enterprise installs.

### DIFF
--- a/modules/bigeye/cloudfront.tf
+++ b/modules/bigeye/cloudfront.tf
@@ -1,0 +1,51 @@
+module "cloudfront" {
+  source = "terraform-aws-modules/cloudfront/aws"
+  count  = var.cloudfront_enabled ? 1 : 0
+
+  aliases             = [local.vanity_dns_name]
+  comment             = local.name
+  enabled             = true
+  is_ipv6_enabled     = true
+  price_class         = "PriceClass_All"
+  http_version        = "http2and3"
+  wait_for_deployment = false
+  logging_config = {
+    bucket = var.cloudfront_logging_bucket
+    prefix = "bigeye.com"
+  }
+
+  origin = {
+    bigeye = {
+      domain_name         = module.haproxy.dns_name
+      connection_attempts = 3
+      connection_timeout  = 10
+      custom_origin_config = {
+        http_port                = 80
+        https_port               = 443
+        origin_keepalive_timeout = 5
+        origin_protocol_policy   = "https-only"
+        origin_read_timeout      = 30
+        origin_ssl_protocols     = ["TLSv1.2"]
+      }
+    }
+  }
+
+  default_cache_behavior = {
+    target_origin_id           = "bigeye"
+    viewer_protocol_policy     = "redirect-to-https"
+    compress                   = var.cloudfront_compression_enabled
+    use_forwarded_values       = false
+    cache_policy_name          = "UseOriginCacheControlHeaders"
+    origin_request_policy_name = "Managed-AllViewer"
+
+    allowed_methods = ["DELETE", "GET", "HEAD", "OPTIONS", "PATCH", "POST", "PUT"]
+    cached_methods  = ["GET", "HEAD", "OPTIONS"]
+  }
+
+  viewer_certificate = {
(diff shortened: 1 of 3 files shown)


modules/bigeye/cloudfront.tf
```diff
@@ -0,0 +1,166 @@
...
+    # includes headers, cookies, and query strings from the viewer request
+    AllViewer = "216adef6-5c7f-47e4-b989-5492eafa07d3",
+  }
+
+  cloudfront_managed_cache_policies = {
+    CachingOptimized             = "658327ea-f89d-4fab-a63d-7e88639e58f6",
+    UseOriginCacheControlHeaders = "83da9c7e-98b4-4e11-a168-04f0df8e2c65",
+  }
+
+}
```
roman-bigeye, line 12: I think it is better to use data sources with for_each for these instead of hardcoding them as local vars.

modules/bigeye/cloudfront.tf
```diff
@@ -0,0 +1,166 @@
...
+
+resource "aws_cloudfront_distribution" "this" {
+  count           = var.cloudfront_options.enabled && var.cloudfront_options.logging_bucket != "" && var.cloudfront_options.acm_certificate_arn != "" ? 1 : 0
+  enabled         = true
```
roman-bigeye, line 16: We can't assume that logging bucket already exists, cert probably too.
Instead of not creating entire distribution, we should use dynamic blocks that are affected by those variables and put null as values if they are not specified.

modules/bigeye/main.tf
```diff
@@ -427,8 +427,8 @@ resource "aws_route53_record" "apex" {
   name    = local.vanity_dns_name
   type    = "A"
   alias {
-    name                   = module.haproxy.dns_name
-    zone_id                = module.haproxy.zone_id
+    name                   = var.cloudfront_options.enabled ? aws_cloudfront_distribution.this[0].domain_name : module.haproxy.dns_name
+    zone_id                = var.cloudfront_options.enabled ? aws_cloudfront_distribution.this[0].hosted_zone_id : module.haproxy.zone_id
```
roman-bigeye, line 431: var.cloudfront_enabled looks like a better condition here.

modules/bigeye/variables.tf
```diff
@@ -2552,3 +2552,30 @@ variable "lineageplus_solr_instance_type" {
...
 }
+
+#======================================================
+# Cloudfront Variables
+#======================================================
+variable "cloudfront_enabled" {
+  description = "It can be advanteous to put a CDN in front of the UI to improve response times to distant regions/zones"
+  type        = bool
+  default     = false
+}
```
roman-bigeye, line 2563: looks like this var is not used

roman-bigeye (changes requested):
Have you considered using [community module](https://github.com/terraform-aws-modules/terraform-aws-cloudfront) instead of implementing all of this? 

It is already well thought through and will require much less maintenance on our part going forward.

modules/bigeye/cloudfront.tf
deeno35: I guess another way of saying what I am going for is "not going to implement unless someone asks for it".  Right now I just don't see any of our self-hosted installers wanting to use Cloudfront

deeno35:
> Have you considered using [community module](https://github.com/terraform-aws-modules/terraform-aws-cloudfront) instead of implementing all of this?
> 
> It is already well thought through and will require much less maintenance on our part going forward.

I looked at it and could not understand why they made it at all.  It's a wrapper that doesn't look like it even streamlines the configuration much at all.  The options track just about 1:1 in the example

modules/bigeye/variables.tf
deeno35: Ah yeah this is dangling.  I wanted to talk to you about using a block config instead of loose vars before going further

roman-bigeye (approved):

PR #414, opened 2025-02-07
fix: add byovpc support for aws_service_discovery_private_dns_namespace

roman-bigeye:


modules/bigeye/main.tf
```diff
@@ -2977,6 +2977,6 @@ resource "aws_appautoscaling_policy" "internalapi_request_count_per_target" {
 
 resource "aws_service_discovery_private_dns_namespace" "this" {
   name = "${local.name}.internal"
-  vpc  = module.vpc[0].vpc_id
+  vpc  = length(var.byovpc_vpc_id) > 0 ? var.byovpc_vpc_id : module.vpc[0].vpc_id
```
deeno35, line 2980: lovely byovpc.

I feel like we should leverage the local.create_vpc var here instead of duplicating the logic

deeno35 (approved):
lovely.  byovpc

modules/bigeye/main.tf
roman-bigeye: yeah, we already check that, replaces with `local.vpc_id`

PR #426, opened 2025-02-21
feat: add new lineageplus_solr variables

roman-bigeye:
Added new vars:
  - `var.lineageplus_solr_opts` - allows to specify any arguments to be passed to the solr startup script
  -  `var.lineageplus_solr_desired_count` - allows stopping solr without deleting the service.

### DIFF
--- a/modules/bigeye/lineageplus.tf
+++ b/modules/bigeye/lineageplus.tf
@@ -24,4 +24,6 @@ module "lineageplus_solr" {
   image_registry   = local.image_registry
   image_repository = "solr${var.lineageplus_solr_image_repository_suffix}"
   image_tag        = var.lineageplus_solr_image_tag
+  solr_opts        = var.lineageplus_solr_opts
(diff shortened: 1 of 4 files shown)


modules/bigeye/lineageplus.tf
```diff
@@ -24,4 +24,6 @@ module "lineageplus_solr" {
   image_registry   = local.image_registry
   image_repository = "solr${var.lineageplus_solr_image_repository_suffix}"
   image_tag        = var.lineageplus_solr_image_tag
+  solr_opts        = var.lineageplus_solr_opts
+  number_of_tasks  = var.lineageplus_solr_number_of_tasks
```
deeno35, line 28: "desired_count" is the standard ECS naming for this

deeno35 (approved):

modules/bigeye/lineageplus.tf
roman-bigeye: renamed

PR #434, opened 2025-02-22
feat: make solr heap configurable via variable

roman-bigeye:


modules/solr-single-instance/main.tf
```diff
@@ -235,7 +235,7 @@ resource "aws_ecs_task_definition" "solr" {
       environment : [
         { name : "SOLR_HOME", "value" : "/var/solr/configs" },
         { name : "SOLR_DATA_HOME", "value" : "/var/solr/data" },
-        { name : "SOLR_HEAP", "value" : "4g" },
+        { name : "SOLR_HEAP", "value" : var.solr_heap_size },
```
deeno35, line 238: I feel like we should generate this, or it will become a maintenance item

deeno35 (approved):

modules/solr-single-instance/main.tf
roman-bigeye: updated to set to 80% of instance memory if not explicitly specified.

PR #443, opened 2025-03-07
refactor: move route53 records management in simpleservice module

roman-bigeye:
This will unify DNS CNAME management and will reduce the amount of copy-pasted resources required to add a new service.


deeno35 (approved):

deeno35:
It will be interesting to see what refactor ends up as.  I would assume a patch version, but I guess we'll see

PR #453, opened 2025-03-19
feat: send solr logs to firelens and jmx stats to datadog

deeno35:


modules/solr-single-instance/containers.tf
```diff
@@ -0,0 +1,179 @@
...
+        hostPort      = 8126
+        protocol      = "tcp"
+      },
+      {
+        containerPort = 8125
+        hostPort      = 8125
+        protocol      = "tcp"
+      }
+    ]
+    # Can be uncommented to get debugging logs on console for this container
```
deeno35, line 130: These are spammy and not much value except when trying to change things.  Things just "work" otherwise and these are wasting $

modules/solr-single-instance/containers.tf
```diff
@@ -0,0 +1,179 @@
...
+    mountPoints  = []
+    volumesFrom  = []
+    portMappings = []
+    firelensConfiguration = {
+      type = "fluentbit",
+      options = {
+        "enable-ecs-log-metadata" : "true"
+      }
+    }
+    # Can be uncommented to get debugging logs on console for this container
```
deeno35, line 159: These are spammy and not much value except when trying to change things.  Things just "work" otherwise and these are wasting $

modules/solr-single-instance/containers.tf
```diff
@@ -0,0 +1,179 @@
+#==============================================
```
deeno35, line 1: Mostly cut and paste from simpleserverice/containers.tf

modules/solr-single-instance/containers.tf
```diff
@@ -0,0 +1,179 @@
...
+      SOLR_DATA_HOME = "/var/solr/data"
+      SOLR_HEAP      = length(var.solr_heap_size) > 0 ? var.solr_heap_size : "${local.solr_default_heap_size}M"
+      SOLR_PORT      = tostring(var.solr_traffic_port)
+      SOLR_OPTS      = join(" ", concat(local.solr_default_opts, var.solr_opts))
+  }) : { Name = k, Value = v }]
+  container_environment_secrets = [for k, v in merge(local.datadog_service_secret_arns, var.secret_arns) : { Name = k, ValueFrom = v }]
+
+  primary_container_definition = {
+    name = var.name,
+    # CPU is left to float
```
deeno35, line 20: This is a change.  I am still following with "let the CPU float", but used the memory flag (hard limit) instead of the memoryReservation (soft limit) flag since we are sharing the ECS task with 2 side card containers now and a floating reservation here will lead to a difficult time troubleshooting if there is contention for mem

modules/solr-single-instance/main.tf
```diff
@@ -258,60 +256,17 @@ locals {
...
+  network_mode          = "awsvpc"
   requires_compatibilities = [
     "EC2",
   ]
   runtime_platform {
     cpu_architecture        = "X86_64"
     operating_system_family = "LINUX"
   }
-  execution_role_arn = "arn:aws:iam::${data.aws_caller_identity.this.account_id}:role/ecsTaskExecutionRole"
+  execution_role_arn = var.execution_role_arn
```
deeno35, line 269: We need the stack's role to access secrets, this follows simpleservice

modules/solr-single-instance/containers.tf
```diff
@@ -0,0 +1,179 @@
...
+      options = {
+        "awslogs-group"         = var.cloudwatch_log_group_name
+        "awslogs-region"        = data.aws_region.current.name
+        "awslogs-stream-prefix" = var.name
+      }
+    }
+  }
+
+  datadog_agent_environment_vars = {
+    ECS_FARGATE                    = "true"
```
roman-bigeye, line 71: I don't know how `ECS_FARGATE` env. var is used, but we don't use fargate for solr.

roman-bigeye (approved):

modules/solr-single-instance/containers.tf
deeno35: Good one, I'll remove it

deeno35: I took a quick look at the docs, this is actually necessary and tells the datadog agent to collect fargate stats, so that's what this is about.  Still good to double check it was necessary.

https://docs.datadoghq.com/integrations/ecs_fargate/?tab=webui

PR #458, opened 2025-03-20
feat: add optional CNAME records for solr service

roman-bigeye:


modules/bigeye/lineageplus.tf
```diff
@@ -16,6 +16,7 @@ module "lineageplus_solr" {
   acm_certificate_arn = local.acm_certificate_arn
   dns_name            = var.create_dns_records ? local.lineageplus_solr_dns_name : ""
   route53_zone_id     = var.create_dns_records ? data.aws_route53_zone.this[0].id : ""
+  solr_cnames         = var.lineageplus_solr_cnames
```
deeno35, line 19: This whole area should probably respect the var.create_dns_records var (just looking at the lines right above this.

deeno35 (approved):

modules/bigeye/lineageplus.tf
roman-bigeye: fixed

PR #472, opened 2025-03-24
fix: pin solr CPU reservation for primary container

deeno35:
With this floating, the CPU charts look very strange.  This is an experiement to normalize the CPU charts so it looks like all of the other ECS tasks where 100% full utilization.

Right now due to the floating CPU reservation, the CPU util observed can be 300%, 700% etc.

Imaging showing .5 as the ECS task's CPU reservation
image

### DIFF
--- a/modules/solr-single-instance/containers.tf
+++ b/modules/solr-single-instance/containers.tf
@@ -7,9 +7,12 @@ locals {
   ec2_mem_usable   = data.aws_ec2_instance_type.this.memory_size - local.os_mem_overhead
   container_memory = local.ec2_mem_usable - (var.datadog_agent_enabled ? var.datadog_agent_memory : 0) - (var.awsfirelens_enabled ? var.awsfirelens_memory : 0)
   # default = 80% of container mem to leave headroom for OS etc.
-  solr_heap_size  = length(var.solr_heap_size) > 0 ? var.solr_heap_size : ceil(local.container_memory * 0.8)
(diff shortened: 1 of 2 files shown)


modules/solr-single-instance/containers.tf
```diff
@@ -7,9 +7,12 @@ locals {
   ec2_mem_usable   = data.aws_ec2_instance_type.this.memory_size - local.os_mem_overhead
   container_memory = local.ec2_mem_usable - (var.datadog_agent_enabled ? var.datadog_agent_memory : 0) - (var.awsfirelens_enabled ? var.awsfirelens_memory : 0)
   # default = 80% of container mem to leave headroom for OS etc.
-  solr_heap_size  = length(var.solr_heap_size) > 0 ? var.solr_heap_size : ceil(local.container_memory * 0.8)
-  container_image = "${var.image_registry}/${var.image_repository}:${var.image_tag}"
+  solr_heap_size = length(var.solr_heap_size) > 0 ? var.solr_heap_size : ceil(local.container_memory * 0.8)
+
+  ec2_cpu_units = data.aws_ec2_instance_type.this.default_vcpus * 1024
+  container_cpu = local.ec2_cpu_units - (var.datadog_agent_enabled ? var.datadog_agent_memory : 0) - (var.awsfirelens_enabled ? var.awsfirelens_memory : 0)
```
roman-bigeye, line 14: subtracting datadog_agent_memory from cpu_units does not look right.

roman-bigeye (approved):

PR #482, opened 2025-04-14
feat: add a dedicated general purpose internal load balancer

deeno35:
No actual services are being migrated through the new ALB in this PR.

This is the first step to consolidating the internal load balancers into 1 and using path based routing instead of load balancer per service.

This is a cost savings step as each LB is ~$23/mo and it adds up with so many internal services.

Future PRs will migrate the services over and remove the dedicated per-service ALBs.

### DIFF
--- a/modules/bigeye/locals.tf
+++ b/modules/bigeye/locals.tf
@@ -30,6 +30,8 @@ locals {
   elasticache_subnet_group_name   = local.create_vpc ? module.vpc[0].elasticache_subnet_group_name : var.byovpc_redis_subnet_group_name
   rabbitmq_subnet_group_ids       = local.create_vpc ? module.vpc[0].elasticache_subnets : var.byovpc_rabbitmq_subnet_ids
 
+  internal_alb_ingress_cidrs = concat([var.vpc_cidr_block], var.internal_additional_ingress_cidrs)
+
   # Temporal Task Queues
   catalog_indexing_temporal_queues = ["indexing.v1"]
   root_cause_temporal_queues       = ["issue-root-cause"]
--- a/modules/bigeye/main.tf
+++ b/modules/bigeye/main.tf
@@ -846,6 +846,63 @@ resource "aws_s3_bucket_policy" "large_payload" {
 }
 
 #======================================================
+# ALBs
+#======================================================
+resource "aws_security_group" "internal_alb" {
+  count       = var.create_security_groups ? 1 : 0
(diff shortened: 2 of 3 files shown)


modules/bigeye/locals.tf
```diff
@@ -30,6 +30,8 @@ locals {
   elasticache_subnet_group_name   = local.create_vpc ? module.vpc[0].elasticache_subnet_group_name : var.byovpc_redis_subnet_group_name
   rabbitmq_subnet_group_ids       = local.create_vpc ? module.vpc[0].elasticache_subnets : var.byovpc_rabbitmq_subnet_ids
 
+  internal_alb_ingress_cidrs = concat([var.vpc_cidr_block], var.internal_additional_ingress_cidrs)
```
deeno35, line 33: This would allow someone to for example make calls to the internal load balancer from a monitoring system that is outside of Bigeye, but still on the same network.

modules/bigeye/variables.tf
```diff
@@ -88,6 +88,12 @@ variable "internal_additional_ingress_cidrs" {
   default     = []
 }
 
+variable "internal_extra_security_group_ids" {
+  description = "Additional security group ids to attach to internal load balancers"
```
deeno35, line 92: This is to support customers using "byo VPC" setups and is also a convenient way to add access to specific external service by SG instead of by IP cidr

modules/bigeye/locals.tf
esmioley: Is this like our datadog synthetics?

esmioley (approved):

modules/bigeye/main.tf
```diff
@@ -845,6 +845,63 @@ resource "aws_s3_bucket_policy" "large_payload" {
   })
 }
 
+#======================================================
+# ALBs
+#======================================================
+resource "aws_security_group" "internal_alb" {
+  count       = var.create_security_groups ? 1 : 0
```
esmioley, line 852: mainly because I don't know, I assume `var.create_security_groups` already exists somewhere.

modules/bigeye/locals.tf
deeno35: Datadog synthetics would be coming in from outside the network, so should not have access to system internals (or the internal LB).  

A datadog agent though that would for example poll Temporal's CLI and emit those timeseries to datadog would be an example

PR #485, opened 2025-04-15
fix: move CD failure notification to its own job

deeno35:
This follows our other deployments.  The benefit here is that when jobs are added in the future, it will be more obvious to not forget to add it to the failure notification job and thus silently fail, which was happening with the publish to s3 step.

Closes SRE-4764


.github/workflows/test_and_publish.yaml
```diff
@@ -154,8 +154,21 @@ jobs:
                 }
               ]
             }
-      - name: failure - notify
+      - name: failure - delete branch and tag
```
deeno35, line 157: this looks weird as a review, but the "slack" step was cut and pasted into its own "notify-on-fail" job below.

.github/workflows/test_and_publish.yaml
```diff
@@ -55,15 +55,35 @@ jobs:
     - name: Terraform lint check
       run: terraform -chdir=${{ matrix.directory }} fmt -check
 
+  changelog:
```
deeno35, line 58: This was lifted out of the publish job to be its own job so the output can be used across other jobs (ie the new notify jobs)

mhalverson (approved):

PR #489, opened 2025-04-16
feat: add flag to control TEMPORAL_LARGE_PAYLOAD_ENABLED env var

deeno35:
This will enable/disable the large pyalod converter for java based services.

For customers, basically don't set this, it's a feature flag for rolling out a new feature.  We will be updating the defaults here when the feature is stable and you will get it naturally then.


modules/bigeye/main.tf
```diff
@@ -2251,6 +2251,7 @@ locals {
     TEMPORAL_TARGET                            = "${local.temporal_dns_name}:${local.temporal_lb_port}"
     TEMPORAL_NAMESPACE                         = var.temporal_namespace
     TEMPORAL_SSL_HOSTNAME_VERIFICATION_ENABLED = var.temporal_use_default_certificates ? "false" : "true"
+    TEMPORAL_LARGE_PAYLOAD_ENABLED             = var.datawatch_temporal_large_payload_enabled
```
deeno35, line 2254: This is being set in for the "datawatch_common_env_vars", so all datawatch based services will get it.

TheRedBerrys (approved):

PR #490, opened 2025-04-17
fix: ec2 ecs mem (solr) increments based on cpu size

deeno35:
There are requirements for ECS mem

https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-cpu-memory-error.html

### DIFF
--- a/modules/solr-single-instance/containers.tf
+++ b/modules/solr-single-instance/containers.tf
@@ -3,7 +3,15 @@
 #==============================================
 locals {
   # 1024 was arrived at through experimentation on a m5.xlarge EC2 instance.  768 would not allow a task to start due to no ECS task being large enough
-  os_mem_overhead  = 1.0 * 1024
+  # note on mem increments
+  #   - 16cpu ECS mem must increment in 8GB increments.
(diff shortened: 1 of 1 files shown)


modules/solr-single-instance/containers.tf
```diff
@@ -3,7 +3,15 @@
 #==============================================
 locals {
   # 1024 was arrived at through experimentation on a m5.xlarge EC2 instance.  768 would not allow a task to start due to no ECS task being large enough
-  os_mem_overhead  = 1.0 * 1024
+  # note on mem increments
+  #   - 16cpu ECS mem must increment in 8GB increments.
+  #   - 8cpu ECS mem must increment in 4GB increments
+  #   - 4cpu and below, mem must increment in 1GB increments
+  # We use this when subtracting an overhead amount from the host OS to arrive at the ECS task mem.
+  os_mem_overhead = (
```
deeno35, line 11: There are no case statements in terraform.  A nested ternary is as good as it goes for something that has more than 2 possible states.

catworlddomination (approved):

PR #493, opened 2025-04-24
feat!: add dedicated lineage api service

deeno35:
This adds a new service - lineageapi.  Its purpose will be to service backend calls for lineage related endpoints.

BREAKING CHANGE: as part of adding a new service, a previous change to security group rules had to be cleaned up.  No action by installers is required other than to install and upgrade to the previous major version first before installing this one.

Note that this PR does not handle routing of API calls, that will be an app version change (TBD).  This PR only handles installation of the service.  By default there will be an instance count of 0 so no costs will be incurred until the feature goes live.


modules/bigeye/main.tf
```diff
@@ -3063,6 +3076,139 @@ resource "aws_appautoscaling_policy" "internalapi_request_count_per_target" {
...
+  centralized_lb_https_listener_rule_arn = aws_lb_listener.https_internal.arn
+  healthcheck_path                       = "/health"
+  healthcheck_grace_period               = 300
+  ssl_policy                             = var.alb_ssl_policy
+  acm_certificate_arn                    = local.acm_certificate_arn
+  lb_idle_timeout                        = 900
+  lb_subnet_ids                          = local.internal_service_alb_subnet_ids
+  lb_additional_security_group_ids       = concat(var.lineageapi_lb_extra_security_group_ids, [module.bigeye_admin.client_security_group_id])
+  lb_additional_ingress_cidrs            = var.internal_additional_ingress_cidrs
+  lb_deregistration_delay                = 90
```
deeno35, line 3113: Note this setting.  The assumption is that this reasonable requests need to be serviced in far shorter than this time.  Losing huge "unfortunate" requests during scale-in and deployments is acceptable.

modules/bigeye/variables.tf
```diff
@@ -2474,6 +2474,90 @@ variable "internalapi_autoscaling_config" {
...
+variable "lineageapi_cpu" {
+  description = "Amount of CPU to allocate"
+  type        = number
+  default     = 1024
+}
+
+variable "lineageapi_memory" {
+  description = "Amount of Memory in MB to allocate"
+  type        = number
+  default     = 8192
```
deeno35, line 2502: This is double the usual base service memory.  Lineage calls can require large amounts of memory.

modules/admin/main.tf
```diff
@@ -49,6 +50,7 @@ locals {
     METRICWORK_ELB_NAME   = var.metricwork_resource_name
     ROOTCAUSE_ELB_NAME    = var.rootcause_resource_name
     INTERNALAPI_ELB_NAME  = var.internalapi_resource_name
+    LINEAGEPI_ELB_NAME    = var.lineageapi_resource_name
```
mhalverson, line 53: typo, should be API not PI

modules/admin/variables.tf
```diff
@@ -139,6 +139,11 @@ variable "internalapi_domain_name" {
   type        = string
 }
 
+variable "lineageapi_domain_name" {
+  description = "lineageapiapi domain name"
```
mhalverson, line 143: typo, too many `api`s

modules/alarms/variables.tf
```diff
@@ -2597,6 +2705,42 @@ variable "ecs_internalapi_mem_threshold" {
   default     = 70
 }
 
+variable "ecs_lineageapi_mem_disabled" {
+  description = "Whether to disable the specific alarm"
+  type        = bool
+  default     = true
+}
+
+variable "ecs_lineageapi_mem_dataponts_to_alarm" {
```
mhalverson, line 2714: typo `dataponts`

mhalverson (approved):

PR #499, opened 2025-04-24
fix: pin sha for all github actions and fix publish GH action

deeno35:
commit 7e956b4d6788688f1206224b7acfe2fec4818cd9 (HEAD -> david/sre-4859-terraform-modules-publish-version-failed-in-gh-action, origin/david/sre-4859-terraform-modules-publish-version-failed-in-gh-action)
Author: David Nguyen <david@bigeye.com>
Date:   Thu Apr 24 15:09:32 2025 -0700

    fix: fix busted publish GH action

    This action isn't working after https://github.com/bigeyedata/terraform-modules/pull/485/files

    The changelog needs git checkout.  Overall the premise of running
    the changelog in its own step was completely unncessary, so I
    put it back to fix the error and have used the output from within
    the publish step to drive the notification jobs instead.

commit b50fc511803b7c76976a174eb9575544ecdfc972
Author: David Nguyen <david@bigeye.com>
Date:   Thu Apr 24 14:53:29 2025 -0700

    fix: pin sha for all github actions

    The test and publish action is failing, might as well handle this
    now since the actions all have to be tested anyways.

    Pinning the sha is a security enhancement over using tagged versions
    as the tags are not immutable and can be retagged onto different
    commits.  Most deterministic is using the SHA (this is GH's
    recommendation as well)

PR #501, opened 2025-04-24
fix: update slack action formatting

deeno35:
This bug was introduced in https://github.com/bigeyedata/terraform-modules/pull/499

The slack action upgrade was a major semver update which required a different interface


mhalverson (commented):
:shipitparrot:

mhalverson (approved):

PR #509, opened 2025-05-02
feat: increase solr os ebs volume size and fix ebs tags

deeno35:
The OS root EBS volume can get uncomfortably full 
at 30GB because solr logs are quite high volume.

We are also missing tags on the OS EBS volume

### DIFF
--- a/modules/solr-single-instance/main.tf
+++ b/modules/solr-single-instance/main.tf
@@ -59,6 +59,22 @@ resource "aws_launch_template" "solr" {
   metadata_options {
     http_tokens = "required"
   }
+
+  ebs_optimized = true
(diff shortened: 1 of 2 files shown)


modules/solr-single-instance/main.tf
```diff
@@ -59,6 +59,22 @@ resource "aws_launch_template" "solr" {
   metadata_options {
     http_tokens = "required"
   }
+
+  ebs_optimized = true
+  block_device_mappings {
+    ebs {
+      encrypted   = "true"
```
deeno35, line 66: We want to use encrypted disks everywhere.

catworlddomination (approved):

PR #511, opened 2025-05-06
feat: add dedicated LB for external calls into Bigeye - part 1

deeno35:
This will mirror the pattern used for the internal LB which is a single LB with path based routing instead of LB per service.

Overall this is a cost savings architecture improvement and does not improve performance/scalability outside of being more thrifty with internal IPs that each LB uses.


deeno35:
superceded by https://github.com/bigeyedata/terraform-modules/pull/551

PR #519, opened 2025-05-27
feat: add datawatch lineageplus config

deeno35:
This is plumbing environment variables through to the app for Datawatch to make calls to the lineage API service.

### DIFF
--- a/modules/bigeye/main.tf
+++ b/modules/bigeye/main.tf
@@ -2276,6 +2276,10 @@ locals {
 
     USE_KMS    = local.using_kms
     KMS_KEY_ID = local.using_kms ? local.kms_key_id : ""
+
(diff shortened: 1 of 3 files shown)


modules/bigeye/main.tf
```diff
@@ -2276,6 +2276,10 @@ locals {
 
     USE_KMS    = local.using_kms
     KMS_KEY_ID = local.using_kms ? local.kms_key_id : ""
+
+    MC_BASE_URL    = "https://${local.base_dns_alias}-metacenter-portal.${var.top_level_dns_name}/portal"
```
deeno35, line 2280: double check on this address.  Basically is the protocol and adding /portal to the path appropriate?

Also be aware this will route through an internal load balancer at some point in the near future.  Right now there is only an external public LB for LineagePlus.

deeno35: Actually I forgot we can use the -internal address too.  Updated.

PR #521, opened 2025-05-29
fix: update MC username

TheRedBerrys:
MetaCenter has an encryption scheme that means we have to use the encrypted version rather than the normal version. There is no feasible way to get the encrypted version from the normal version right now, except manually to replace it as we are doing here.

In the future, we hope to update MetaCenter to accept the decrypted version of the username from Datawatch, but right now we can't.


deeno35:
The linting error is actually about formatting of the git commit.  I'm surprised you were able to push a commit actually.

The format of the subject should be

fix: update MC username

PR #522, opened 2025-05-29
fix: update MC username

TheRedBerrys:
It didn't let me fix the commit message in #521 so this is another attempt.


deeno35 (approved):
Looks great!  I think you won't mind, but I'm going to merge this to speed things along

PR #528, opened 2025-05-30
fix: fix missing launch template ebs device name

deeno35:
This is a bug introduced in https://github.com/bigeyedata/terraform-modules/pull/509

Without this, the autoscaling group fails to refresh and leaves solr in a down state.


akatt1082 (approved):
Approved. Change looks minimal

PR #534, opened 2025-06-13
feat: add CPU autoscaling for datawatch

deeno35:
This can be disabled or set to target tracking, but is set to cpu_scaling by default.  30% has been a reliable balance between cost and slow scale out time and overall helps the system cope with the occasionally overly large request that can come in.

Note the service will be replaced if you don't intervene and this will cause an outage

Run this before running terraform apply to avoid the service replacement. tf  state mv 'module.bigeye.module.datawatch.aws_ecs_service.controlled_count[0]' \ 'module.bigeye.module.datawatch.aws_ecs_service.uncontrolled_count[0]'


TheRedBerrys (approved):
Looks like good default values to me.

PR #535, opened 2025-06-14
feat: allow enabling alb anomaly mitigation

deeno35:
Anomaly mitigation is a newer routing feature on ALB which ideally solves both hot spots due to large requests as well as avoiding the case where a sick instance (like with a broken DB connection) can get a large percentage of inbound requests routed to it because it processes them very quickly and thus has the "least number of active connections".

This isn't available in conjunction with session stickiness so is not used on the frontend, but will set this everywhere else.

BREAKING CHANGE: The aws provider version upgrade is required in order to control the enable flag for this feature.

### DIFF
--- a/modules/bigeye/main.tf
+++ b/modules/bigeye/main.tf
@@ -801,6 +801,7 @@ resource "aws_s3_bucket_lifecycle_configuration" "models" {
     expiration {
       days = 45
     }
+    filter {}
   }
 }
 
@@ -833,6 +834,7 @@ resource "aws_s3_bucket_lifecycle_configuration" "large_payload" {
     expiration {
       days = 7
     }
+    filter {}
   }
 }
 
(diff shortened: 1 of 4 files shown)


modules/bigeye/main.tf
```diff
@@ -987,18 +989,19 @@ module "haproxy" {
...
+  healthcheck_unhealthy_threshold   = 3
+  ssl_policy                        = var.alb_ssl_policy
+  acm_certificate_arn               = local.acm_certificate_arn
+  lb_idle_timeout                   = var.lb_timeout
+  lb_subnet_ids                     = var.internet_facing ? local.public_alb_subnet_ids : local.internal_service_alb_subnet_ids
+  lb_additional_security_group_ids  = concat(var.haproxy_lb_extra_security_group_ids, [module.bigeye_admin.client_security_group_id])
+  lb_additional_ingress_cidrs       = var.additional_ingress_cidrs
+  lb_stickiness_enabled             = true
+  lb_deregistration_delay           = 900
+  load_balancing_anomaly_mitigation = false
```
esmioley, line 1004: Just want to confirm this is deliberately false for haproxy and web

esmioley (approved):

PR #552, opened 2025-07-02
feat: enable EC AZ rebalancing and LB anomaly mitigation by default

deeno35:
This will set the default for these features to on.  The flags to control this will then be removed sometime down the road.

Closes SRE-5313 SRE-5166


mhalverson (approved):
>This will set the default for these features to no.

I think you mean "on"

PR #554, opened 2025-07-08
feat: add flag to enable additional debugging stats for java services

deeno35:
This will enable stats such as jdbc connection pool utiliazation to be emitted to Datadog.


jfcole (approved):
💸

PR #692, opened 2026-05-14
[SRE-5189] Phase 1a: Add optional static-assets bucket to bigeye module

n0t0:


modules/bigeye/locals.tf
```diff
@@ -205,7 +205,10 @@ locals {
       "type" : "mcp-gateway"
       "retention_days" : 30
     }
-  ]
+    ], var.static_asset_bucket_enabled ? [{
+      "type" : "static-assets"
+      "retention_days" : 7
```
esmioley: Do we want this to be longer? A deploy with no hotfixes could exceed 7 days by hours, and sometimes we skip deploys (thanksgiving week, holiday week)?

esmioley (approved):

PR #694, opened 2026-05-15
[SRE-5189] Phase 2a: Add optional S3 origin for static-asset Cloudfront cache behaviors

n0t0:


jylertones (approved):
I kind of understand what this is doing

PR #704, opened 2026-06-04
[SRE-6999] Pin datawatch RDS sql_mode in module defaults

n0t0:
## Why
  Today `sql_mode` is not set anywhere in our Terraform, so MySQL falls back to whatever the engine ships as default. On the `8.0.40` and `8.4.9` engines we currently run, that default is:

  `ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION`

We've been running with `ONLY_FULL_GROUP_BY` enabled for the entire life of these stacks - neither the Java app nor Python services reference `sql_mode` anywhere (verified via grep across `java/`, `python/`, all yaml/properties/JDBC URLs). Per Backend Weekly (SRE-6999), we want to lock this in explicitly so:
  - A future engine-default change (e.g. the upcoming 8.0 → 8.4 rollout, or any later bump) can't silently shift sql_mode under us.
  - The primary and replica parameter groups stay in sync.

### DIFF
--- a/modules/bigeye/variables.tf
+++ b/modules/bigeye/variables.tf
@@ -1937,6 +1937,10 @@ variable "datawatch_rds_default_parameters" {
       value        = 1
       apply_method = "pending-reboot"
     }
+    sql_mode = {
+      value        = "ONLY_FULL_GROUP_BY,STRICT_TRANS_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION"
+      apply_method = "immediate"
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


mhalverson:
> We've been running with ONLY_FULL_GROUP_BY enabled for the entire life of these stacks

No, it's been *disabled* the entire time

This is the mystery to me since Mysql has enabled it by default since 5.7.5

I'm not comfortable with this change, we want to try enabling it on staging only for 2 weeks

n0t0:
See: https://github.com/bigeyedata/semantic-data-platform/pull/27502 ; Closing.

